PR #12, opened 2022-04-19
Improve build infrastructure

lionel-:
- Use makefiles to drive the build from inside and outside the docker environment.

- The makefiles have no knowledge of the source and target files inside each sub-build. So we use dummy targets in a `state` folder to record completion of steps and to allow reset.

  For instance, to recompile, relink, and reinstall R after updating the patches, type `rm state/r-stage2 && make`. In the future, we could add specialised targets to only rebuild parts of R, e.g. `src/main` or `src/library/base`.

- Use configurable variables instead of hardcoded paths. By defaults, all paths are relative to the webR root folder so that running `make` does not create side effects elsewhere on the file system.

- The wasm libs are installed in `ROOT/wasm`

- The native tools and libs (e.g. dragonegg) are installed in `ROOT/host`

- The `scripts` folder is now `tools` and contains things like libgfortran and dragonegg.

- The relevant parts of libgfortran are now embedded instead of downloading all of gcc.

- The base URL in `webR.js` is configured from the `BASE_URL` envvar at build time, which can also be set (like other build variables) from `~/webr-config.mk`.

In an upcoming PR, I'll look into adding support for building Fortran files using our custom flang branch.


Makefile
```diff
@@ -1,9 +1,16 @@
...
 	docker build -t webr-build .
-	docker run --rm --mount type=bind,source=$(PWD)/dist,target=/app/webR webr-build
+	docker run --rm --mount type=bind,source=$(PWD),target=/app webr-build
 	cp loader/repl.html dist/index.html
 	cp loader/webR.js dist/webR.js
 
+.PHONY: container-%
+container-%:
+	docker build -t webr-build .
+	docker run -dit --name $* --mount type=bind,source=$(PWD),target=/app webr-build bash
```
georgestagg, line 12: I think this is supposed to start an interactive `bash` inside the docker image? However, when I run `make container-webr` it starts the docker process but does not drop me into the shell.

Can we change it to,

```
docker run -it --rm --name $* --mount type=bind,source=$(PWD),target=/app webr-build bash
```
which seems to work better for me? The extra `--rm` also just makes docker clean up after itself so the command can run again next time.

R/Makefile
```diff
@@ -0,0 +1,223 @@
...
+
+$(BUILD)/state/r-stage2: $(BUILD)/state/r-stage1 $(BUILD)/state/r-stage2-configured
+# Remove lazy loaded base package so that R can cleanly rebuild
+	rm -f "$(R_HOST)/lib/R/library/base/R/base"*
+	( \
+	  cd $(R_BUILD)/src/library/base; \
+	  $(MAKE_WASM) Rsimple; \
+	)
+	mkdir -p "$(R_WASM)/lib/R/library/base/R"
+	cp -r "$(R_SOURCE)/build/library/base/R/." "$(R_WASM)/lib/R/library/base/R"
```
georgestagg, line 152: I think lines 151 and 152 should be just:
```
cp -r "$(R_SOURCE)/build/library/base/R/." "$(R_HOST)/lib/R/library/base/R"
```
That is, first removing base from `$(R_HOST)/lib/R/library/` and then copying back the one built by using Rsimple into the same place.

The intention here is to remove the lazy loaded version of the base package from the host's version of R. That way when we later run the stage2 R build process setting `R_EXE` to the host's R, it can happily perform the lazy loading process on the base package, rather than complaining that it is already lazy loaded.

R/Makefile
```diff
@@ -0,0 +1,223 @@
...
+	)
+# Rebuild R.bin.js and R.bin.data
+	( \
+	  cd $(R_BUILD); \
+	  $(MAKE_WASM) R; \
+	  $(MAKE_WASM) install-wasm; \
+	)
+	cp -r "$(R_WASM)/dist/." $(DIST)
+	touch $@
+
```
georgestagg, line 173: We should probably add a `clean` target to this makefile that removes the `BUILD` and `DOWNLOAD` directories so the process can be easily restarted.

R/Makefile
georgestagg: Note that without this change, I see an error during building because R_EXE can't find the base package (...because we quite rudely deleted it from its library).

georgestagg:
This looks like an excellent improvement to the build, thank you so much! In general I agree that your changes are sensible ones, particularly removing the strict dependence on hard coded file paths and allowing for local overrides with `webr-config.mk`

I have tried the new process on my own machine and have just a few comments, which I've made directly on the particular lines above.

After tweaking things on my side, I was able to get a working build of webR, so I'll be more than happy to merge this once they above questions have been resolved.

Makefile
lionel-: With a non-detached `--rm`, we get dropped into the container shell, but exiting causes the container to be deleted automatically. The intention is to create a persistent container where work can be interrupted and resumed as needed. I don't mind attaching it automatically. Would this make it slightly less clear that this make target creates persistent state though?

R/Makefile
lionel-: yeah sorry this was the case initially and I've already changed it back locally. I got a little confused while trying to solve a local build issue in my flang branch.

R/Makefile
lionel-: Good idea. I've also split the toplevel `clean` target into `clean` and `distclean`. The former only removes build artifacts and preserves the `dist` folder.

R/Makefile
lionel-: This should now be back to normal.

Makefile
georgestagg: Right. No worries, I had just misunderstood the purpose of the target. I am so used to just creating docker containers and blowing them away without worry that I had not considered persistence. I'm happy to leave it as it is.

The only other thing I'm a little worried about is bind mounting everything at `/app`. I had originally wanted only the `dist` directory bind mounted, and everything else made temporary inside the docker container so that when I delete the docker container only the final build remains and temporary files are gone. However, that problem has been mostly solved by makefiles now anyway (with a neat option of override), so it does not really matter.

R/Makefile
georgestagg: Great, thanks!

georgestagg:
I've just tried this once again and it builds great. I'm happy to merge.

PR #27, opened 2022-05-19
Add `rebuild-main` target

lionel-:
This new target quickly rebuilds any changed `.c` file in `src/main` and recreates the distributed wasm file. Helpful for quick iterative development.


lionel-:
With latest commits, also rebuild `src/unix` since there are many webR changes in that folder. The target is now called `rebuild-core` to reflect that files outside of `src/main` are rebuilt.

PR #29, opened 2022-05-29
Fix webR in a worker on Safari

georgestagg:
Disabling use of bcEval seems to lead to less call stack being used,
helping for Safari where at the time of writing the default amount of
stack allocated to worker threads is only 512KB, compared to 8MB on
the main thread.


lionel-:
oh that makes sense. `bcEval()` also causes compilation with sanitizers to stall for a very long time.

georgestagg:
Merged as part of #30

PR #31, opened 2022-06-28
Make optimisation flags configurable

lionel-:
`R.bin.js` is minified since this change: https://github.com/georgestagg/webR/commit/619830c4c28b94542c2450cb7bb460f023388997. This slows down the link step and makes it harder to inspect the Emscripten JS code while debugging.

This PR adds an optimisation flag for the link step and make both flags configurable via `~/.webr-config.mk`.

### DIFF
--- a/R/Makefile
+++ b/R/Makefile
@@ -34,7 +34,8 @@ WASM_LIBS = $(PCRE_WASM_LIB) $(XZ_WASM_LIB) $(FORTRAN_WASM_LIB)
 # Configure your local environment in this file
 -include ~/.webr-config.mk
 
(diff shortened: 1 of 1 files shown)


R/Makefile
```diff
@@ -34,7 +34,8 @@ WASM_LIBS = $(PCRE_WASM_LIB) $(XZ_WASM_LIB) $(FORTRAN_WASM_LIB)
 # Configure your local environment in this file
 -include ~/.webr-config.mk
 
-WASM_OPT := -Oz
+WASM_OPT ?= -Oz
```
georgestagg, line 37: 👀

PR #35, opened 2022-07-05
Add support for local webr-repo configuration

lionel-:
The variables `WEBR_REPO` and `WEBR_LIB` can now be configured in `~/.webr-config.mk`. These should point to the `lib` and `repo` folders created with https://github.com/lionel-/webr-repo/. This is mainly for local dev and debugging.

- `WEBR_REPO` specifies a local CRAN-like repository. Then `webr_install()` from https://github.com/lionel-/webr-repo/blob/main/install.R can be used to install packages.

- `WEBR_LIB` specifies a local directory of built packages. These are automatically installed.

The PR also disables `HAVE_UMASK` to fix a permission error triggered within the internal `untar()` method called by `webr_install()`.

### DIFF
--- a/R/Makefile
+++ b/R/Makefile
@@ -33,6 +33,8 @@ WASM_LIBS = $(PCRE_WASM_LIB) $(XZ_WASM_LIB) $(FORTRAN_WASM_LIB)
 
 # Configure your local environment in this file
 -include ~/.webr-config.mk
+export WEBR_REPO
+export WEBR_LIB
 
 WASM_OPT ?= -Oz
 WASM_OPT_LDADD ?= $(WASM_OPT)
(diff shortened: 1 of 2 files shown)


R/Makefile
```diff
@@ -144,6 +146,9 @@ $(BUILD)/state/r-stage2-configured: $(BUILD)/state/r-patched $(WASM_LIBS)
 	      --enable-byte-compiled-packages=no \
 	      --enable-static=yes \
 	      --host=wasm32-unknown-emscripten
+# Disable umask which doesn't work well within Emscripten. Fixes
+# permission issues when extracting tarballs.
+	sed -i "" "/#define HAVE_UMASK/d" $(R_SOURCE)/build/src/include/config.h
```
georgestagg, line 151: Sorry to be picky, but we have a few similar tweaks in `patches/R-4.1.3/emscripten-unimplemented.diff` and we should probably be consistent.

I think I'd prefer it as an ifdef guarded patch rather than running `sed`. Is there a reasonable header file or C file in the R source we could patch, adding something like
```
#ifdef __EMSCRIPTEN__
#undef HAVE_UMASK # Disable umask which doesn't work well within Emscripten.
#endif
```
instead?

georgestagg: OK. That's a good argument. I remember having some `sed` lines to make similar changes in the past, but removed them. I had convinced myself that the patches were better but I forget my exact arguments. Possibly it was before the more flexible Makefile system you brought in. I like that it means we can switch up versions of R without having to worry about those specific patches.

For consistency I'd like to also move the `undef`'s out of `emscripten-unimplmented` into `Makefile`, then. I think almost everything in there is global and can be moved, and I'm happy to merge and deal with that myself.

PR #37, opened 2022-07-08
Switch to blocking approach

lionel-:
This PR switches webR to a blocking approach. Whenever R is asking for input through the `ReadConsole()` platform API point, we block in the worker thread until an input is available.

- We no longer use Comlink to communicate between threads. All communication is now done through a special channel data structure that reads and writes messages.

- In the main thread, a request-response may be initiated. This returns an awaitable promise that resolves with the resonse. The worker thread can only read and write inputs at the moment. If needed we could add a public request-response method later on.

- The `main.c` patches are now much lighter. We can directly call into the regular REPL from `importScripts()`, and never return. And we can now use nested REPLs such as `browser()` or `utils::menu()`.

- I reorganised the files a bit. All the synchronisation code is in the `chan` subfolder. We now have `webr-main.js` and `webr-worker.js` files.

Synclink implementation:

- Internally, the worker-side channel uses a private and synchronous request-response method to request input messages. The implementation is directly taken from the synclink project (https://github.com/hoodmane/synclink).

- While integrating Synclink in webR, I realised that it would not be difficult to use only the parts that interest us (in particular we don't need all the proxying abstractions). When the project stabilises, we should be able to request adjustments so that we don't need to maintain all this synchronisation code in webR.

- The Synclink code is currently mostly unchanged. However I plan to add support for `reviver` and `replacer` methods in the JSON serialisation routine that is used to communicate responses to a blocked worker. Currently, passing back data structures like `Uint8Array` is a little awkward. The custom reviver and replacer methods will handle these cases. This should eventually be contributed upstream so we can use Synclink or its successor in the future.


src/package-lock.json
```diff
@@ -1,6 +1,5001 @@
 {
```
lionel-, line 1: I must have done something wrong here, with these 5k locs... I'll take a look.

georgestagg: It could just be `npm` weirdness? If so, you could try to rebase and separate the change into its own commit?

lionel-: I see this when I run `npm install`, it's probably the explanation:

```
npm WARN old lockfile
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile
npm WARN old lockfile This is a one-time fix-up, please be patient...
npm WARN old lockfile
```

I've removed all the extra diffs from this PR so that we can update the lock file later on.

PR #42, opened 2022-07-21
Fix loading webR cross-domain

georgestagg:
With this commit, the main webR class and the JS worker can be loaded into web pages on different origins.

The commit adds a workaround that replaces using `loadScripts(url)` and `Worker(url)` directly with fetching the file content with XHR first, creating a URL blob with the content, then loading that. That way, the content is allowed to run even when the webR files and the source content are loaded via different origins.

The workaround kicks in only if the configured `BASE_URL` begins with the string `http`. Otherwise it is assumed loading the URL directly is OK.

The commit also edits our esbuild config to build some of our outputs as "neutral" type `.mjs` files, rather than using the "node" platform for everything. The switch to neutral will be needed later for loading webR as a JS module into a page from a CDN URL. The switch to using the `.mjs` extension is not strictly required for the browser, but it is required when running in node with this platform format.

### DIFF
--- a/src/esbuild.js
+++ b/src/esbuild.js
@@ -1,7 +1,7 @@
 const esbuild = require('esbuild');
 const cssModulesPlugin = require('esbuild-css-modules-plugin');
 
-function build({ input, output, minify }) {
+function build({ input, output, platform, minify }) {
   esbuild.build({
     entryPoints: [`${input}`],
     bundle: true,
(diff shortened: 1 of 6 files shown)


lionel- (commented):
I don't understand why or how the XHR requests work around the cross-origin rules but the changes sound reasonable. It'd be nice to add comments above the XHR paths to explain what's going on.

georgestagg:
Thanks, I've added a comment explaining what is happening.

PR #47, opened 2022-08-17
eval_R_4: `evalRCode` patch to R

georgestagg:
The `evalRCode` R patch has been reworked to use `R_ParseEvalString` and `R_TryCatch`. The result of running the provided code and the content of any thrown error are stored in a `VECSXP` and returned to webR.

Later, webR will inspect the contents of the first two elements of the returned vector to get the resulting `SEXP` object, and check for thrown errors e.g. "parse error".

A finally function for the try catch is setup, and its returned `SEXP` included in the vector returned to webR, but the body is currently empty.

It is possible to test `webR.evalRCode()` here, but it will just return a pointer to the R `VECSXP` in WASM memory.


patches/R-4.1.3/eval-r-code.diff
```diff
@@ -0,0 +1,64 @@
...
++#include <emscripten.h>
++#include <emscripten/fetch.h>
++
++/* TODO: This patch should be moved to an external emscripten SIDE_MODULE
++ * This will help to keep webR features separate from required patches to R core
++ */
++typedef struct {
++    const char *str;
++    SEXP env;
++} eval_args;
```
lionel-, line 19: In rlang and vctrs we just use the full `struct foo` type, as in the Linux kernel style guide. Up to you which style to use in webR's C code, but I think typedef'd types should have a `_t` suffix to help readability.

lionel- (approved):

PR #48, opened 2022-08-17
eval_R_5: The RObj JS interface

georgestagg:
The JS interface for various types of R objects are implemented in this PR.

Includes some new getter methods, based on calling R's built in get symbols:
* `get` - R's `[`
* `get2` - R's `[[`
* `getDollar` - R's `$`.

Since we're using R's operators, indexing starts from 1. This should probably be looked at further and a decision discussed on if we'd like to shift things to start at 0 in JS.

It is possible to test the interface using this PR, but it is a bit fiddly. You either need to hack the source of the worker thread to add code to test things, or you can do it in browser dev tools if you switch to the worker thread and then be sure to pause execution.

e.g.

```
> r = await webR.evalRCode('c(1,2,3)');
< 18524104
[... switch to the worker thread in devtools ...]
[... pause the currently blocking worker thread in devtools ...]
> r = RObj.wrap(18524104)
> r.get2(1).toJs() // The first element of the vector returned by evalRCode is the resulting object
< Float64Array(3) [1, 2, 3, buffer: ...]
```

Things are probably easier to test in the next PR where the main thread proxies are introduced.


src/webR/robj.ts
```diff
@@ -0,0 +1,501 @@
...
+    let idx: RPtr;
+    let char: RPtr = 0;
+    if (typeof prop === 'number') {
+      idx = Module._Rf_protect(Module._Rf_ScalarInteger(prop));
+    } else {
+      char = allocateUTF8(prop);
+      idx = Module._Rf_protect(Module._Rf_mkString(char));
+    }
+    const call = Module._Rf_protect(Module._Rf_lang3(RObj.bracketSymbol.ptr, this.ptr, idx));
+    const sub = RObj.wrap(Module._Rf_eval(call, RObj.globalEnv.ptr));
```
lionel-, line 125: I would evaluate in the baseenv just in case `[` gets masked by the search path.

In general, evaluation of internal code in the globalenv is unhygienic. Instead we should try to reproduce the semantics of namespaced code (imports -> base namespace -> search path --- the last one being an implementation detail of scoping in R). One easy way to approximate this is to evaluate in the baseenv (the last environment before empty-env).

src/webR/robj.ts
```diff
@@ -0,0 +1,501 @@
...
+    let idx: RPtr;
+    let char: RPtr = 0;
+    if (typeof prop === 'number') {
+      idx = Module._Rf_protect(Module._Rf_ScalarInteger(prop));
+    } else {
+      char = allocateUTF8(prop);
+      idx = Module._Rf_protect(Module._Rf_mkString(char));
+    }
+    const call = Module._Rf_protect(Module._Rf_lang3(RObj.bracket2Symbol.ptr, this.ptr, idx));
+    const sub = RObj.wrap(Module._Rf_eval(call, RObj.globalEnv.ptr));
```
lionel-, line 141: Same regarding base env.

src/webR/robj.ts
```diff
@@ -0,0 +1,501 @@
...
+    Module._Rf_unprotect(2);
+    if (char) Module._free(char);
+    return sub;
+  }
+
+  getDollar(prop: string): RObj {
+    const char = allocateUTF8(prop);
+    const idx = Module._Rf_protect(Module._Rf_mkString(char));
+    const call = Module._Rf_protect(Module._Rf_lang3(RObj.dollarSymbol.ptr, this.ptr, idx));
+    const sub = RObj.wrap(Module._Rf_eval(call, RObj.globalEnv.ptr));
```
lionel-, line 151: Same regarding base env.

src/webR/robj.ts
```diff
@@ -0,0 +1,501 @@
...
+  }
+
+  static unprotect(n: number): void {
+    Module._Rf_unprotect(n);
+  }
+
+  static unprotectPtr(obj: RObj): void {
+    Module._Rf_unprotect_ptr(obj.ptr);
+  }
+}
```
lionel-, line 205: It would be useful to also expose `R_PreserveObject()` and `R_ReleaseObject()`. These will be needed for automatic protection for instance.

src/webR/robj.ts
```diff
@@ -0,0 +1,501 @@
...
+    return RObj.wrap(Module._INTERNAL(this.ptr));
+  }
+}
+
+class RObjPairlist extends RObj {
+  toObject(): { [key: string]: RawType } {
+    const d: { [key: string]: RawType } = {};
+    for (let next = this as Nullable<RObjPairlist>; !next.isNull(); next = next.cdr()) {
+      const symbol = next.tag();
+      if (!symbol.isNull() && symbol.printname()) {
```
lionel-, line 247: When is an `RObjString` falsy?

src/webR/robj.ts
```diff
@@ -0,0 +1,501 @@
...
+        d[symbol.printname().toJs()] = next.car().toJs();
+      } else {
+        d[Object.keys(d).length] = next.car().toJs();
+      }
+    }
+    return d;
+  }
+
+  includes(name: string): boolean {
+    return !this.get2(name).isNull();
```
lionel-, line 257: This is incorrect since `includes('bar')` will return `NULL` for `quote(foo(bar = NULL))`. Should this be checking for `undefined`?

I think it's problematic to be introducing such a large exposed API in one go without unit tests. Perhaps you could already prepare a set of tests in the form of console strings and their output? This could then be turned into proper tests later on.

src/webR/robj.ts
```diff
@@ -0,0 +1,501 @@
...
+}
+
+class RObjString extends RObj {
+  toJs(): string {
+    return UTF8ToString(Module._R_CHAR(this.ptr));
+  }
+}
+
+class RObjEnv extends RObj {
+  ls(): string[] {
```
lionel-, line 328: Should be `names()` since `all.names` defaults to `TRUE`.

src/webR/robj.ts
```diff
@@ -0,0 +1,501 @@
...
+class RObjLogical extends RObjAtomicVector {
+  toArray(): Int32Array {
+    return Module.HEAP32.subarray(
+      Module._LOGICAL(this.ptr) / 4,
+      Module._LOGICAL(this.ptr) / 4 + this.length
+    );
+  }
+  getIndex(idx: number): RLogical {
+    const elem = this.toArray()[idx];
+    if (typeof elem === 'undefined') return undefined;
```
lionel-, line 408:
```suggestion
    if (typeof elem === 'undefined') {
      return undefined;
    }
```

lionel-: This seems strange to my R programmer eyes. Shoudn't we throw an error when indexing atomic vectors out of bounds, consistently with R?

lionel- (approved):
Looks great!

Regarding the names of subsetting methods, how about:

- `get()` for `[[`
- `subset()` or `slice()` (my preference) for `[`


> Since we're using R's operators, indexing starts from 1. This should probably be looked at further and a decision discussed on if we'd like to shift things to start at 0 in JS.

I'd be tempted to just pass indices straight to R, so that we get full indexing semantics (e.g. 0 for empty objects, negative indices, etc).

src/webR/robj.ts
```diff
@@ -0,0 +1,501 @@
...
+
+  includes(name: string) {
+    return !this.getDollar(name).isUnbound();
+  }
+
+  get(prop: number | string): RObj {
+    if (typeof prop === 'number') {
+      throw new Error('Object of type environment is not subsettable');
+    }
+    return this.getDollar(prop);
```
lionel-, line 344: A nice thing about evaluating a `$` call is that it solves complexities such as what happens with promises. Perhaps mention this in a comment?

src/webR/robj.ts
georgestagg: Whoops! That test was left over from a time when `printname()` returned a JS string. I've removed `&& symbol.printname()` from the test; even for symbols with strange print names like `R_MissingArg` I don't think the result is ever falsy.

src/webR/robj.ts
georgestagg: I've reworked this, I think it should now work when the items are `NULL`.

I'll keep a note to include this in the list of proper unit tests later.

src/webR/robj.ts
georgestagg: I've removed this line. In fact, I've removed `getIndex`. I never intended that particular function to be used as part of the public API anyway, just as a helper for `toJs()`. Rather than making it private, I've just merged the `getIndex` and `toJs` functions where relevant.

The issue of out of bounds access should not come up in `toJs()`, as those methods work aware of the length of the `Module.HEAP` subarray and indexing from 0.

PR #54, opened 2022-08-24
Fix R signrank and wilcox functions

georgestagg:
Currently functions such as `qwilcox` and `dsignrank` throw a
`RuntimeError`, caused by a WASM function signature mismatch.

Emscripten does not deal with function pointer casting well, so
this commit adds an unused `SEXP args` argument to the functions
`stats_signrank_free` and `stats_wilcox_free`, located in the default
stats library, to help work around this.

### DIFF
--- a/patches/R-4.1.3/fix-signrank-wilcox.diff
+++ b/patches/R-4.1.3/fix-signrank-wilcox.diff
@@ -1,0 +1,32 @@
+Index: R-4.1.3/src/library/stats/src/distn.c
+===================================================================
+--- R-4.1.3.orig/src/library/stats/src/distn.c
++++ R-4.1.3/src/library/stats/src/distn.c
+@@ -458,13 +458,13 @@ DEFMATH4_2(qtukey)
+ extern void signrank_free(void);
+ extern void wilcox_free(void);
+ 
+-SEXP stats_signrank_free(void)
++SEXP stats_signrank_free(SEXP args)
+ {
+     signrank_free();
+     return R_NilValue;
+ }
+ 
+-SEXP stats_wilcox_free(void)
++SEXP stats_wilcox_free(SEXP args)
+ {
+     wilcox_free();
+     return R_NilValue;
+Index: R-4.1.3/src/library/stats/src/statsR.h
+===================================================================
+--- R-4.1.3.orig/src/library/stats/src/statsR.h
++++ R-4.1.3/src/library/stats/src/statsR.h
+@@ -203,5 +203,5 @@ SEXP Fisher_sim(SEXP sr, SEXP sc, SEXP s
+ SEXP chisq_sim(SEXP sr, SEXP sc, SEXP sB, SEXP E);
+ SEXP d2x2xk(SEXP sK, SEXP sm, SEXP sn, SEXP st, SEXP srn);
+ 
+-SEXP stats_signrank_free(void);
+-SEXP stats_wilcox_free(void);
++SEXP stats_signrank_free(SEXP args);
++SEXP stats_wilcox_free(SEXP args);
--- a/patches/R-4.1.3/series
+++ b/patches/R-4.1.3/series
@@ -15,3 +15,4 @@ tre-fix-rf_error-args.diff
 emscripten-disable-bcEval.diff
 emscripten-xhr-download.diff
 eval-r-code.diff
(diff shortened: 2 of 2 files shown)


patches/R-4.1.3/fix-signrank-wilcox.diff
```diff
@@ -0,0 +1,32 @@
+Index: R-4.1.3/src/library/stats/src/distn.c
+===================================================================
+--- R-4.1.3.orig/src/library/stats/src/distn.c
++++ R-4.1.3/src/library/stats/src/distn.c
+@@ -458,13 +458,13 @@ DEFMATH4_2(qtukey)
+ extern void signrank_free(void);
+ extern void wilcox_free(void);
+ 
+-SEXP stats_signrank_free(void)
++SEXP stats_signrank_free(SEXP args)
```
lionel-, line 10: I see this implementation:

```c
SEXP stats_signrank_free(void)
```

And this declaration:

```c
void signrank_free(void);
```

So I'm a bit surprised that we change the arguments rather than the return type?

lionel- (commented):
It would be nice to send a patch to fix in R. I sent a similar patch to fix another mismatch a while ago: https://bugs.r-project.org/show_bug.cgi?id=18321

patches/R-4.1.3/fix-signrank-wilcox.diff
georgestagg: There's two functions of interest for each problem,

in `src/nmath/signrank.c`:
```
void signrank_free(void)
{
    w_free();
}
```

and in `src/library/stats/src/distn.c`:

```
SEXP stats_signrank_free(void)
{
    signrank_free();
    return R_NilValue;
}
```

So, when we invoke `stats_signrank_free` it just calls `signrank_free()`.

Now, in `src/library/stats/src/init.c` we have the following two lines as part of an array of .External methods.

```
    {"signrank_free", (DL_FUNC) &stats_signrank_free, 0},
    {"wilcox_free", (DL_FUNC) &stats_wilcox_free, 0},
```

Emscripten doesn't like the function pointer cast here, but I noticed that it does not mind the cast just a couple of lines above, referencing `call_dqags`, which has definition:
```
SEXP call_dqags(SEXP args)
```

So, after noticing that, I changed both the definition and implementation of `stats_signrank_free` to take a list of arguments: `SEXP stats_signrank_free(SEXP args)`, but left `signrank_free` alone.

With that change, if I understand things correctly, when `.External(signrank_free)` is run in R, `stats_signrank_free` is invoked passing an empty list of arguments. Emscripten is happy, and `stats_signrank_free` is happy because it just ignores the arguments passed in and calls `signrank_free()` without any args, and that's OK because we have not touched that one.

Then the same thing for `stats_wilcox_free` and `wilcox_free`.

lionel-: oops sorry I was confused. Does this mean we'll run into trouble for every zero-argument C callable? If so that doesn't seem sustainable.

What error does Emscripten emit? Is this a bug in Emscripten? Could we adjust the `DL_FUNC` typedef to something that Emscripten accepts?

georgestagg: I think this is still a bug in R, my reading of https://cran.r-project.org/doc/manuals/R-exts.html#Calling-_002eExternal is that the function signature *should* be `SEXP stats_signrank_free(SEXP args)`, where `args` is populated a list (I assume empty in this case). So we should only see trouble where that has not been done, and I think we should be able to upstream this patch in future.

Emscripten compiles OK, but it aborts at runtime with `Uncaught (in promise) RuntimeError: null function or function signature mismatch`. Presumably because WASM is looking for a `stats_signrank_free` that takes in a single argument.

I think this works OK outside of WASM and not yet caught because everything is just cast to `DL_FUNC`. Function pointer casts is a known issue in Emscripten (https://emscripten.org/docs/porting/guidelines/function_pointer_issues.html), but I've not had much luck with the `EMULATE_FUNCTION_POINTER_CASTS` workaround in the past.

lionel-: Ah now I understand, I was thinking in terms of the `.Call` interface rather than `.External` (the former is more common in newer packages). Thanks for the explanation.

PR #55, opened 2022-08-24
WIP: Unit Tests

georgestagg:
Note: this branch contains a copy of 5ec84e5 that is already merged into main, and the Wilcox/signrank function fix submitted separately in #54.

These changes adds a new `make check` target that runs tests under webR running in Node. The webR API is tested via `.js` files under the `tests` directory, and R scripts are tested running in webR via the `testRCode.js` file and R scripts located in the `tests/R` directory.

The tests are orchestrated by Makefiles that run the JS/R scripts in webR and log the output. The output is compared to "expected" output provided in `out.save` files. Deviation from the expected output is considered a test failure.

**TODO:** Some test output depends on machine architecture, the current time, and other randomness, and should be removed or ignored.

The provided R test scripts are based on copies of the files created by `tools::testInstalledPackage`
for each of R's default packages. Test scripts have been modified to
remove obviously broken tests, such as those that depend on `help` content
or other files that are not packaged as part of the webR virtual
filesystem (e.g. `COPYING`, `TESTING`, `NEWS`, etc.)

I have only looked at the `testInstalledPackage` script output a little, so I cannot say for sure that the expected output provided is actually correct. However, by adding the Wilcox fix and removing a test that calls `chol(m, pivot = TRUE))`, (leading to an unrecoverable runtime error in Fortran), these tests at least seem to run cleanly now. That is, even if they are incorrect they do not cause the webR WASM runtime to crash.


tests/evalRCode.js
```diff
@@ -0,0 +1,235 @@
...
+	await webR.init();
+  let r;
+
+    try {
+    // RObjNull
+    r = await webR.evalRCode('NULL');
+    console.log(await r.toString());
+    console.log(await r.type);
+    console.log(await r.isNull());
+    console.log(await r.toJs());
```
lionel-, line 24: Basing tests only (or mostly) on output comparison is not good practice. Is this common in JS? I would have expected most or all of these to be equality tests.

lionel- (commented):
There are two big features in this PR, testing base packages (a) and testing webR itself (b).


Regarding (a), I'm not sure I understand where do the test files come from? Since running R CMD check is not possible because most of the tests are run in subprocesses we should come up with a strategy that doesn't pin tests to a particular version but at the same time requires minimal maintenance efforts when switching to a newer R version. If possible, we should pick the files to test straight from installed packages. If modifications are required, we'd apply them in patches at installation time.

I think the two most important things to check are examples and tests. For both examples and tests, the main thing is that they run without errors and we can generally ignore the outputs. Sometimes packages have saved `.Rout` files in their tests folder and then it makes sense to also check the output, if that is not too difficult or cumbersome.


Regarding (b), it would be very helpful to embed a test library that would provide a range of test predicates for snapshot and equality comparisons, and most importantly handle reporting when a test fails.

georgestagg:
Going forward I'll think about this some more and split this into two PRs for a) and b).

PR #57, opened 2022-08-26
Add a minimal webr R package to the webR build process

georgestagg:
The package is installed into the webR filesystem as part of the stage-2 R build.

Currently the package only contains a single function `install`, which can be used to install packages from a webR binary CRAN-like repo.

In the future I expect more will be added to extend the functionality of webR, or to expose webR interfaces to the R code running within.


packages/webr/R/install.R
```diff
@@ -0,0 +1,56 @@
+#' Install one or more packages from a webR binary package repo
+#'
+#' @export
+#'
+#' @param packages character vector containing the names of packages to install
+#' @param repos character vector containing the URIs of the webR repos to use
+#' @param lib the library directory where the packages will be installed
```
lionel-, line 7: Should be in sentence case with full stops.

lionel- (approved):

PR #58, opened 2022-08-26
Add installPackages function to webR and download R packages automatically in the REPL

georgestagg:
Includes #57.

The `installPackages` function calls the `webr::install` R function internally, using the configured `PKG_URL` as the binary repo. 

With this, R packages are downloaded automatically in the webR REPL when requested by running a command containing calls to `library` or `require`.


lionel- (commented):
I would prefer to stick to manual installation with `webR::install()` instead of intercepting `library()` and `require()` calls on the JS side, which is a bit hacky and not 100% reliable.

The other thing we could do is to install a global calling handler for `packageNotFoundError` that prompts the user about installing the required package with `install()`. Then we call the `retry_loadNamespace` restart if it exists. I think there are cases where the this doesn't work but it's been on my todo list to fix upstream (IIRC the classed errors are sometimes swallowed up by intervening handlers).

georgestagg:
> I would prefer to stick to manual installation with `webR::install()` instead of intercepting `library()` and `require()` calls on the JS side, which is a bit hacky and not 100% reliable.

OK. The idea here is to make things as smooth as possible for someone who has just found webR and is trying out the demo REPL for the first time. It would be nice to match the automatic package downloading feature in Pyodide's demo at https://pyodide.org/en/stable/console.html when you run something like `import numpy as np`. Although IIRC they use a less hacky method to detect if a package should be downloaded, I'll check.

I'll look into hooking `packageNotFoundError` and prompting the user if they'd like to download the package. I think this would still be a pretty smooth process for new users. Perhaps it would also be nice if the prompt is only given when the package requested actually exists in the default configured webR binary repo.

georgestagg:
For the moment, I've tweaked the way packages are automatically installed.

As part of the startup of the demo REPL app (but not in the core `WebR` startup), a handler is installed for `packageNotFoundError` that prompts the user asking if they would like to try to download a webR binary package.

If the user replies in the negative, the package error is thrown as it would normally.

If the user replies in the affirmative, the package is downloaded using `webr::install()`. The `retry_loadNamespace` restart is invoked if it exists, continuing execution where it left off. Otherwise, the `abort` restart is invoked so as to return to the top level, squelching the missing package error message.

It is a shame, but it looks like `retry_loadNamespace` does not work with `library`. However, commands in the form of `cli::cat_line("Hello, World!")` now work in the REPL, if one replies with  "Yes" when asked if they want to install the package.

---

The older changes adding `installPackages` to the WebR API are no longer strictly required, but I have left them in this PR because I think they will be a useful interface for developers.

src/repl/repl.ts
```diff
@@ -68,6 +69,26 @@ const webR = new WebR({
 (async () => {
   await webR.init();
 
+  webR.writeConsole(`
```
lionel-, line 72: Does this cause an empty prompt to show up on start? It might be better to use `evalRCode()` instead, to avoid this.

Also it's probably better to implement this in the webR package as something like `global_handle_install()`, see https://github.com/r-lib/rlang/blob/main/R/cnd-handlers.R for examples.

georgestagg: I don't see an empty prompt on startup, but I also agree that `evalRCode` would be better. Unfortunately that causes `Error: should not be called with handlers on the stack`. I think it's because `evalRCode` now wraps `R_ParseEvalString` in a `tryCatch`.

I'll migrate the code into the new `webr` package and consider what to do with `evalRCode` tomorrow. Perhaps we can add some flags to control if `R_tryCatch` is used, or just create another function that calls `R_ParseEvalString` directly. Those patches will be moving to a webr package SIDE_MODULE anyway in future, so we can be fairly flexible with them.

lionel-: yup a bare bones wrapper around `R_ParseEvalString()` sounds good.

packages/webr/R/handlers.R
```diff
@@ -0,0 +1,36 @@
+#' Prompt user to install missing packages from a webR binary repo
+#'
+#' @description
+#' When enabled, `packageNotFoundError` thrown causes a menu prompt
```
lionel-, line 4:
```suggestion
#' When enabled, `packageNotFoundError` errors cause a menu prompt
```

packages/webr/R/handlers.R
```diff
@@ -0,0 +1,36 @@
...
+#' to be shown to the user asking if they would like to attempt to
+#' download the missing package.
+#' 
+#' If the user replies in the negative, the package error is thrown as
+#' it would normally. If the user replies in the affirmative, the package
+#' is downloaded using `webr::install()`.
```
lionel-, line 10:
```suggestion
#' is downloaded using [webr::install()].
```

packages/webr/R/handlers.R
```diff
@@ -0,0 +1,36 @@
...
+#' 
+#' If it is possible to do so, execution will continue without
+#' interrupting the current program.
+#' 
+#' @export
+global_prompt_install <- function() {
+  globalCallingHandlers(
+    packageNotFoundError = function(cnd) {
+      if (!interactive()) return()
```
lionel-, line 19:
```suggestion
      if (!interactive()) {
        return()
      }
```

packages/webr/R/handlers.R
```diff
@@ -0,0 +1,36 @@
...
+      pkg <- cnd$package
+      download <- utils::menu(
+        c("Yes", "No"),
+        title=paste(
```
lionel-, line 23:
```suggestion
        title = paste(
```

lionel-: Should probably use `paste0()` here.

packages/webr/R/handlers.R
```diff
@@ -0,0 +1,36 @@
...
+          'Failed to load package "', pkg,
+          '". Do you want to try downloading it from the webR binary repo?',
+          sep=''
```
lionel-, line 26:
```suggestion
          sep = ""
```

src/webR/webr-main.ts
```diff
@@ -67,14 +73,23 @@ export class WebR {
     return (await this.#chan.request({ type: 'getFSNode', data: { path: path } })) as FSNode;
   }
 
-  async evalRCode(code: string, env?: RObject): Promise<RObject> {
-    if (env && !isRObject(env)) {
+  async evalRCode(
+    code: string,
+    options: { env?: RObject; withHandlers?: boolean } = {}
```
lionel-, line 78: Maybe keep `env` as its own argument? It's good for it be more prominent, both for consistency with R's `eval()` functions, and also to remind users they should think about the evaluation environment.

lionel-: I like that `withHandlers` is "buried" in an options dict, since it's such a low-level parameter.

lionel- (approved):

packages/webr/R/handlers.R
georgestagg: Sorry, I keep doing that! It's an old habit...

src/webR/webr-main.ts
georgestagg: I considered that, but I wasn't sure how happy I was with having to pass something like `undefined` or a reference to the global env there just to change something in `options`. It would be nicer with proper named parameters with defaults that you can skip but I think most just use the passing an object trick to emulate it.

I guess the vast majority of the time the `options` won't be touched, so it does not matter much. I like the idea of being consistent with R's `eval()` more.

PR #59, opened 2022-08-26
Alternative implementation of RObj proxies

georgestagg:
This is an alternative to PR #49 and #50.

This alternative implementation of `RProxy` is intended to be a simpler and more intuitive interface to `RObj`s.

With this version of `RProxy`, the proxy is made aware of all valid `RObj` methods at creation time. They use this list to only proxy calls that are valid to the worker thread. Accessing any other property is considered invalid and returns `undefined`.

All calls to `RObj` methods through `RProxy` return promises to the result.

Accessing properties of R objects can no longer be done though proxy get semantics. i.e. you cannot do `proxy[2]` or `proxy.foo` and instead you use `await proxy.get(2)` or `await proxy.pluck('foo')`.

There are no set and apply proxy semantics. i.e. you cannot do `proxy()` or `proxy[2] = 42`. Instead you use `await proxy.exec()` or `await proxy.set(2, 42)`.

This should make it easier to reason about what `RProxy` is doing, and more obvious that one must await the result since all valid properties on the object are functions that return Promises, rather than the previous implementation of recursively returning `RProxy` lazily.

When calling `RProxy` methods, the intended functionality is that wherever you can pass an `RObj` you instead pass an `RProxy`, and wherever an `RObj` is returned, an `RProxy` is returned to the main thread. The proxy `get` hook takes care of the `RObj`<->`RProxy` conversion automatically where required, by passing messages with `RTargetObj` type set to `PTR` for proxies, or `RAW` otherwise. This conversion process should be tested thoroughly in our future planned unit tests.

The type instability of using `RTargetRaw/RTargetPtr` and invoking all methods though the proxy `get` hook is masked from the user. Accurate type information and function signatures for `RProxy` methods are built by mapping `RObj` classes using typescript (see the first half of `proxy.ts` for details).


src/webR/webr-worker.ts
```diff
@@ -125,18 +142,72 @@ function downloadFileContent(URL: string, headers: Array<string> = []): XHRRespo
...
+    obj,
+    Array.from({ length: args.length }, (_, idx) => {
+      const arg = args[idx];
+      return arg.type === RTargetType.PTR ? RObj.wrap(arg.obj) : arg.obj;
+    })
+  ) as RawType | RObj;
+
+  if (isRObj(res)) {
+    return { obj: res.ptr, methods: RObj.getMethods(res), type: RTargetType.PTR };
+  }
```
lionel-, line 177: `else`

src/webR/webr-main.ts
```diff
@@ -64,10 +66,24 @@ export class WebR {
   async getFSNode(path: string): Promise<FSNode> {
     return (await this.#chan.request({ type: 'getFSNode', data: { path: path } })) as FSNode;
   }
-  async evalRCode(code: string): Promise<number> {
-    return (await this.#chan.request({
+
+  async evalRCode(code: string, env?: RProxy<RObj>): Promise<RProxy<RObj>> {
```
lionel-, line 70: I think it'd be better if `RProxy` was a pure implementation detail that is not exposed to users. Would the following scheme work?

- Unexport `RProxy`.
- Rename `RObj` to `RObjWorker`.
- Export this alias for public usage:
  `export type RObj = RProxy<RObjWorker>`

src/webR/robj.ts
```diff
@@ -318,29 +467,34 @@ class RObjList extends RObj {
...
     call.setcar(this);
     let c = call.cdr();
     let i = 0;
     while (!c.isNull()) {
-      c.setcar(args[i++]);
+      c.setcar(argObjs[i++]);
       c = c.cdr();
     }
-    return RObj.wrap(Module._Rf_eval(call.ptr, RObj.globalEnv.ptr));
+    const res = RObj.wrap(Module._Rf_eval(call.ptr, RObj.globalEnv.ptr));
```
lionel-, line 485: Evaluation in the base namespace.

lionel- (approved):
Nice! If we could fully hide `RProxy` from the public API, as suggested in a comment, that'd be perfect.

src/webR/webr-main.ts
georgestagg: Good idea hiding the `RProxy` detail where possible. I don't think we want to entirely un-export, though. It might come in useful for someone and I think I'll need to export it anyway if we want to then export the user-facing type aliases from `robj.ts` (which feels like the right place) rather than `proxy.ts`.

I'm not sure about `RObjWorker`, comparing this with the subclass names one might think this an interface to an R worker thread or something.

How would you feel about the following?

 * `RProxy` remains exported but we pledge to avoid using it in user-facing code such as in `webr-main.ts`.
 * The type alias `export type RObject = RProxy<RObj>` is added for public use.

The name `RObject` has the benefit that it is not abbreviated, matching the other type aliases. I think it also matches Rcpp better.

lionel-: Sounds good regarding exporting `RProxy`. I think we should make it clear this is a mostly internal type though. In R you can export objects marked `internal` which causes them to not appear in the documentation index. I feel like we'd export `RProxy` under this scheme if this was R.

I don't love the idea of having both `RObj` and `RObject`. These names are really close and it isn't clear what the differences are. Also the shorter variant should probably be the user facing one. How about `RObjImpl` on the worker side? These types wouldn't be exported / exposed to end users right?

georgestagg: `RObjImpl` on the worker side works for me. I'll also rename `isRProxy`->`isRObject` for public use, and `isRObj` to `isRObjImpl` for internal use.

Yes, the idea is that end users should not work with `RProxy` or `RObjImpl` directly. However, I think we do still need to export things so that we ourselves can use them in other files running on the worker side.

When we get around to writing docs and adding more JSDoc strings you can mark things as `@private` so they are not included in the output. I'm not 100% sure, but I don't think there's an easy way in typescript to export objects/types but just internally for us. I will do some reading to try and find out for sure. If not, we might have to simply rely on our documentation not including information about these private objects, even if technically they are exported.

lionel-: yup perfect.

PR #60, opened 2022-09-06
Unit tests for webR's Typescript code using the Jest testing framework

georgestagg:
Requires #51 for Node compatibility.

This PR sets up unit testing for Typescript using the Jest framework and `ts-jest`. Code coverage is calculated by setting Node's `NODE_V8_COVERAGE` env var and then using the `c8` package to calculate statistics. This is a required change to Jest's default method for code coverage so that the lines of code that are run in the web worker thread can be tracked and recorded.

The testing has been added to the tasks in the GitHub Actions files, and the resulting code coverage is uploaded to codecov.io. This gives us access to statistics over time, badges, etc: codecov

The PR includes tests for the core webR initialisation and R object code (via `evalRCode`). No tests for the REPL and console apps have been included yet, but should not be difficult to include in the future once the structure of those apps have stabilised.

NOTE: The diff for `package-lock.json` is very large, triggered by adding the `jest` and `c8` node packages. I have tried to keep that change in its own commit, but I'm not sure it can be suppressed completely.

### DIFF
--- a/src/tests/webr.test.ts
+++ b/src/tests/webr.test.ts
@@ -117,17 +117,17 @@ describe('Evaluate R code', () => {
   test('Throw an error if passed an invalid environment', async () => {
     // @ts-expect-error Deliberate type error to test Error thrown
     const promise = webR.evalRCode('3.14159', { env: 42 });
-    await expect(promise).rejects.toThrowError();
+    await expect(promise).rejects.toThrow('invalid environment object');
   });
 
   test('Throw an error if passed an invalid environment object type', async () => {
     const euler = await webR.evalRCode('0.57722');
-    await expect(webR.evalRCode('x', euler)).rejects.toThrow();
+    await expect(webR.evalRCode('x', euler)).rejects.toThrow('env argument with invalid SEXP type');
   });
 
   test('Throw errors from R', async () => {
     const badSyntax = webR.evalRCode('42+');
-    await expect(badSyntax).rejects.toThrow();
+    await expect(badSyntax).rejects.toThrow('parse error');
   });
 
   test('Write to stderr while evaluating R code', async () => {
--- a/src/tests/webr.test.ts
+++ b/src/tests/webr.test.ts
@@ -131,8 +131,10 @@ describe('Evaluate R code', () => {
   });
 
   test('Write to stderr while evaluating R code', async () => {
+    await webR.flush();
     const res = webR.evalRCode('message("Hello, stderr!")');
     await expect(res).resolves.not.toThrow();
+    expect((await webR.read()).data).toBe('Hello, stderr!');
   });
 });
 
--- a/src/tests/webr.test.ts
+++ b/src/tests/webr.test.ts
@@ -53,6 +53,20 @@ describe('Test webR simple console input/output', () => {
   });
 });
 
+describe('Download and install binary webR packages', () => {
+  test('Install packages via evalRCode', async () => {
+    await webR.evalRCode('webr::install("cli", repos="https://repo.webr.workers.dev/")');
+    const pkg = (await webR.evalRCode('"cli" %in% library(cli)')) as RLogical;
+    expect(await pkg.toLogical()).toEqual(true);
+  });
+
+  test('Install packages via API', async () => {
+    await webR.installPackages(['MASS']);
+    const pkg = (await webR.evalRCode('"MASS" %in% library(MASS)')) as RLogical;
+    expect(await pkg.toLogical()).toEqual(true);
+  });
+});
+
 describe('Test webR virtual filesystem', () => {
   const testFileContents = new Uint8Array([1, 2, 4, 7, 11, 16, 22, 29, 37, 46]);
   test('Upload a file to the VFS', async () => {
--- a/src/tests/webr.test.ts
+++ b/src/tests/webr.test.ts
@@ -1,0 +1,369 @@
+/* eslint-disable @typescript-eslint/await-thenable */
+import { WebR } from '../webR/webr-main';
+import { Message } from '../webR/chan/message';
+import {
+  RNull,
+  RDouble,
+  RInteger,
+  RFunction,
+  RSymbol,
+  RLogical,
+  RComplex,
+  RRaw,
+  RPairlist,
+  RList,
+  REnvironment,
+} from '../webR/robj';
+import { sleep } from '../webR/utils';
+
+import util from 'util';
+
+const webR = new WebR({
+  WEBR_URL: '../dist/',
+  RArgs: ['--quiet'],
+});
+
+jest.setTimeout(10000);
+
+describe('Test webR instance startup', () => {
+  test('Constructs successsfully', () => {
+    expect(webR).toBeDefined();
+  });
+
+  test('Initialises successfully', async () => {
+    await expect(webR.init()).resolves.not.toThrow();
+  });
+});
+
+describe('Test webR simple console input/output', () => {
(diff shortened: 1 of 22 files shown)


src/webR/webr-main.ts
```diff
@@ -45,6 +45,10 @@ export class WebR {
     return await this.#chan.initialised;
   }
 
+  terminate() {
```
lionel-, line 48: Maybe `close()` would be more in line with channel terminology?

src/tests/webr.test.ts
```diff
@@ -0,0 +1,407 @@
...
+    }
+    expect(msg.data).toBe('> ');
+  });
+
+  test('Write an R command to the console', () => {
+    expect(() => webR.writeConsole('42\n')).not.toThrow();
+  });
+
+  test('Read result line from stdout', async () => {
+    expect((await webR.read()).data).toBe('[1] 42');
```
lionel-, line 52: It's awesome seeing the REPL come out alive in these tests :)

src/tests/webr.test.ts
```diff
@@ -0,0 +1,407 @@
...
+  });
+});
+
+describe('Download and install binary webR packages', () => {
+  test('Install packages via evalRCode', async () => {
+    await webR.evalRCode('webr::install("cli", repos="https://repo.webr.workers.dev/")');
+    const pkg = (await webR.evalRCode('"cli" %in% library(cli)')) as RLogical;
+    expect(await pkg.toLogical()).toEqual(true);
+  });
```
lionel-, line 61: 👍

src/tests/webr.test.ts
```diff
@@ -0,0 +1,407 @@
...
+    await webR.installPackages(['MASS']);
+    const pkg = (await webR.evalRCode('"MASS" %in% library(MASS)')) as RLogical;
+    expect(await pkg.toLogical()).toEqual(true);
+  });
+});
+
+describe('Test webR virtual filesystem', () => {
+  const testFileContents = new Uint8Array([1, 2, 4, 7, 11, 16, 22, 29, 37, 46]);
+  test('Upload a file to the VFS', async () => {
+    await expect(webR.putFileData('/tmp/testFile', testFileContents)).resolves.not.toThrow();
```
lionel-, line 73: Could also check that R can see the file.

src/tests/webr.test.ts
```diff
@@ -0,0 +1,407 @@
...
+  });
+
+  test('Download a file from the VFS', async () => {
+    const fileContents = await webR.getFileData('/tmp/testFile');
+    expect(fileContents).toStrictEqual(testFileContents);
+  });
+
+  test('Recieve information about a file on the VFS', async () => {
```
lionel-, line 81:
```suggestion
  test('Receive information about a file on the VFS', async () => {
```

lionel-: Also below.

src/tests/webr.test.ts
```diff
@@ -0,0 +1,407 @@
...
+    expect(result._target).toHaveProperty('type', 'PTR');
+    expect(result._target).toHaveProperty('methods');
+    expect(result._target).toHaveProperty('obj');
+    expect(result._target.obj).toEqual(expect.any(Number));
+  });
+
+  test('Throw an error if passed an invalid environment', async () => {
+    // @ts-expect-error Deliberate type error to test Error thrown
+    const promise = webR.evalRCode('3.14159', { env: 42 });
+    await expect(promise).rejects.toThrowError();
```
lionel-, line 118: Is there an easy way to check that the expected error has been thrown and not something else, e.g. by supplying a keyword or part of the message that should match the thrown error?

src/tests/webr.test.ts
```diff
@@ -0,0 +1,407 @@
...
+  });
+
+  test('Throw an error if passed an invalid environment object type', async () => {
+    const euler = await webR.evalRCode('0.57722');
+    await expect(webR.evalRCode('x', euler)).rejects.toThrow();
+  });
+
+  test('Throw errors from R', async () => {
+    const badSyntax = webR.evalRCode('42+');
+    await expect(badSyntax).rejects.toThrow();
```
lionel-, line 128: Nice

src/tests/webr.test.ts
```diff
@@ -0,0 +1,407 @@
...
+  });
+
+  test('Write to stderr while evaluating R code', async () => {
+    const res = webR.evalRCode('message("Hello, stderr!")');
+    await expect(res).resolves.not.toThrow();
```
lionel-, line 133: Also check received output? We might need a `flush()` method to clear inputs before running a new test.

src/tests/webr.test.ts
```diff
@@ -0,0 +1,407 @@
...
+  });
+
+  test('List items in an R environment', async () => {
+    const env = (await webR.evalRCode('x<-new.env();x$a=1;x$b=2;x$.c=3;x')) as REnvironment;
+    let ls = await env.ls();
+    expect(ls).toEqual(['a', 'b']);
+    ls = await env.ls(true);
+    expect(ls).toEqual(['.c', 'a', 'b']);
+    ls = await env.names();
+    expect(ls).toEqual(['.c', 'a', 'b']);
```
lionel-, line 314: 👍

src/tests/webr.test.ts
```diff
@@ -0,0 +1,407 @@
...
+  test('Evaluating R code in an environment', async () => {
+    const env = (await webR.evalRCode('x<-new.env();x$a=1;x$b=2;x$.c=3;x')) as REnvironment;
+    const value = (await webR.evalRCode('b', env)) as RDouble;
+    expect(await value.toNumber()).toEqual(2);
+  });
+});
+
+describe('Invoking RFunction objects', () => {
+  test('Execute an R function with JS arguments', async () => {
+    const choose = (await webR.evalRCode('choose')) as RFunction;
```
lionel-, line 341: Here's an R function I didn't know about!

lionel- (approved):
Looks like a solid set of tests!

Should probably split the test file though. We usually have a one-to-one correspondence between test files and source files, with a strict naming convention. This makes it easy to switch between source and test files in editors.

src/webR/webr-main.ts
georgestagg: I named it after [`Worker.terminate()`](https://developer.mozilla.org/en-US/docs/Web/API/Worker/terminate), which is ultimately what is invoked.

Matching the `read` and `write` terminology has convinced me to rename to `close`, but I'll have to remember to document clearly that `close` is pretty powerful and kills the thread.

lionel-: That makes sense.

src/tests/webr.test.ts
georgestagg: Yes, you can pass a string to match or an error instance to check the right errors are thrown[1].I'll do that and match on parts of the expected error message text.

This was a good suggestion, it revealed a problem transferring Error objects back to the main thread in Safari and Node, even though Error is supposed to support structured cloning. I'll commit a fix as part of these changes.

[1]: https://jestjs.io/docs/expect#tothrowerror

src/tests/webr.test.ts
georgestagg: `flush()` is a great idea! I'll make a small commit as part of this PR adding a `flush()` function to `ChannelMain` that `read()`s until the queue is empty and returns the resulting messages in an array.

Then it should be as simple as `flushing`, running the code, then `read`ing to get the stderr output.

PR #62, opened 2022-09-13
Add option to capture `stdout`, `stderr` and/or conditions when running R code with `evalRcode`

georgestagg:
This PR modifies `evalRCode` so that `stdout` is captured, intended to be as close as possible to how it would be output if the code was run via the REPL. This will use useful for workflows that require evaluating snippets of R code and returning what the output would look like, e.g. online runnable R code examples, without the need to wrap the R REPL with supporting  TS/JS code.

**EDIT**: Note that several statements below are no longer true. Later comments better describe the state of this PR.

---

The `evalRCode` R patch is modified so that an R `textConnection` is constructed before evaluating the code, then hooked up to R's output connection. Once the code has finished the original output connection is restored and the contents of the `textConnection` are passed back in the returned vector as a character vector. IIUC the `textConnection` does not need to be closed, and will be automatically cleaned up by R's GC later.

The R patch has also been modified to take in a `withHandlers` argument, and the webR worker modified to pass a boolean there rather than running `R_ParseEvalString` directly. That way `stdout` can captured both in the case of installing or not installing condition handlers, although not installing will be rare.

The vector returned to webR contains 3 elements: The result, `stdout` lines, and an error condition (or NULL). If there is a non-null error condition, a JS Error is constructed based on the text and thrown (which is caught and then send to the main thread as an object, to be re-thrown there). Otherwise, the entire vector is passed to the main thread. Previously the result was extracted and returned to the main thread alone.

On the main thread, if an error object comes back it is re-thrown. Otherwise, the result object and `stdout` are obtained from the object using `.get()`. The final object as returned to the user has the structure:
```
{
  'result': RObject,
  'stdout': string[],
}
```
with the expectation that over time the vector as it comes from R will be expanded to include SEXPs containing other conditions (e.g. a vector of warnings), and the JS object returned to the user expanded similarly.

Auto-printing, as it is by default in the REPL, does not happen with `R_ParseEvalString`. So, the input code is modified before submission to be `withAutoprint({...}, echo=FALSE)`, with the resulting value from `withAutoprint` returned. Without something like this, I think R code would require modifying to insert `print` statements to match REPL output. It has just now occurred to me that perhaps wrapping with autoprint should be a setting in the `options` argument?

Finally, the tests are updated for the new returned structure and a new test is added for checking that `stdout` with auto-printing works.


georgestagg:
The `evalRCode` patch and webR API has been expanded to allow optional capturing of `stderr` and/or `message` and `warning` conditions. Auto-print has been made optional. The options object passed to `evalRCode` currently has the following properties:

 * `captureStreams` - Should the `stdout` and `stderr` output streams be captured? Default: `false`.
 * `captureConditions` -  Should conditions raised during execution be captured and returned? Default: `true`.
 * `withAutoprint` - Should the code automatically print evaluation output? Default: `false`.
 * `withHandlers` - Should the code be executed using a `tryCatch` with handlers in place? Default: `true`.

The result of the evaluation, `stdout`, `stderr` and `message`/`warning` conditions are returned with the result in an object with constant shape. When not captured, the properties are returned as `null`. If an `error` condition is captured, it is sent to and thrown on the main thread.

These new features and options now makes `evalRCode` very flexible, but more thought will probably still be needed about how to present it to the end user before this can be merged.

georgestagg:
The latest commit in this PR removes the `evalRCode` R patch entirely, instead moving the logic that was introduced there in the earlier PR commits from C code to R code, as a new function in the webr R package.

`R_ParseEvalString` and `Rf_eval` are used to invoke `webr:::evalRCode`, passing in the user code as a character object and the settings provided in the JS `options` object as `T`/`F` arguments.

As with the C patch before, `webr:::evalRCode` executes the user code through a `tryCatch` by default, and so unless the default settings are changed errors are still caught and returned to webR to be thrown.

On the JS the invocation is a bit more complicated in `webR.evalRCode`, but the resulting R code installing the `tryCatch`/`withCallingHandlers` condition handlers and executing the user code is much easier to follow than it was in C, and we no longer need to patch R itself.

georgestagg:
The latest commit combines the various output streams into a single `output` array by means of a custom R connection to be used with `sink`. By using a custom R connection and a combined output the order of events is maintained in the output array.

Types of event in the `output` array are distinguished by the `type` property (`stdout`, `stderr`, `message`, `warning`, `error`).

packages/webr/src/outputconnection.c
```diff
@@ -0,0 +1,101 @@
...
+  data->output = out;
+  data->cur = data->line = data->buf;
+  con->private = data;
+}
+
+SEXP new_output_connections() {
+  Rconnection out_con_stdout;
+  Rconnection out_con_stderr;
+
+  const char *names[] = {"stdout", "stderr", "vec"};
```
lionel-, line 88: Needs a `""` sentinel.

```suggestion
  const char *names[] = { "stdout", "stderr", "vec", "" };
```

packages/webr/src/outputconnection.c
```diff
@@ -0,0 +1,101 @@
...
+    data->buf[BUFSIZE-1] = '\0';
+	  res = BUFSIZE - curlen - 1;
+  }
+  data->cur += res < 0 ? 0 : res;
+
+  // Check for newlines in buffer and copy completed lines to output
+  for (p = data->line; p < data->cur; p++) {
+    if (*p == '\n') {
+      *p = '\0';
+      const char *names[] = {"type", "data"};
```
lionel-, line 49: Needs a `""` sentinel.

```suggestion
      const char *names[] = { "type", "data", "" };
```

src/webR/webr-main.ts
```diff
@@ -126,17 +123,13 @@ export class WebR {
...
-          stderr: await stderr.toJs(),
-          messages: await messages.toJs(),
-          warnings: await warnings.toJs(),
-        };
+        const outList = (await obj.get(2)) as RList;
+        const output = ((await outList.toArray()) as { type: string[]; data: unknown }[]).map(
+          (elem) => {
+            return { type: elem.type[0], data: elem.data };
+          }
+        );
```
lionel-, line 131: Can we wrap this pattern in a helper function?

packages/webr/R/eval.R
```diff
@@ -15,26 +15,23 @@
...
 evalRCode <- function(code, conditions = TRUE, streams = FALSE,
   autoprint = FALSE, handlers = TRUE
 ) {
   res <- NULL
-  errors <- NULL
-  warnings <- NULL
-  messages <- NULL
-  stderr <- NULL
-  stdout <- NULL
+  out <- .Call("new_output_connections")
```
lionel-, line 23: Needs a comment that we can't use individual `textConnection()` for stdout and stderr because we need to multiplex them in a single stream.

lionel-: Prefix C symbols with `ffi_`

```suggestion
  out <- .Call("ffi_new_output_connections")
```

packages/webr/R/eval.R
```diff
@@ -15,26 +15,23 @@
 #' @param handlers If `TRUE`, execute the R code using a [tryCatch], with
 #' handlers in place.
 #'
+#' @useDynLib webr
```
lionel-, line 18: Should use symbol registration

```suggestion
#' @useDynLib webr, .registration = TRUE
```

Need to add an init function to register symbols.

packages/webr/R/eval.R
lionel-: By the way this is doing a symbol search across all dlopened DLLs. Should at least pass the package as argument to avoid this search and prevent unexpected masking. But as argued above, it's better to use symbol registration so that we can `.Call()` into a C function pointer directly.

packages/webr/src/outputconnection.c
```diff
@@ -0,0 +1,101 @@
...
+}
+
+static void output_destroy(Rconnection con) {
+  if(con->private) free(con->private);
+}
+
+static int output_vfprintf(Rconnection con, const char *format, va_list ap) {
+  struct output_con_data* data = con->private;
+  int res;
+  char *p, *q;
```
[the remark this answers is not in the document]
georgestagg: The code was not copied in entirety, but I did learn how the R connections work by reading functions like `dummy_vfprintf`[1] in the R source and basing things on those. Looks like I copied the line truncation warning message at some point and picked up some tabs.

I'll add an attribution comment to the file mentioning `connections.c`, thanks for the reminder.

[1] https://github.com/wch/r-source/blob/trunk/src/main/connections.c#L423

packages/webr/src/outputconnection.c
```diff
@@ -0,0 +1,101 @@
...
+    if (*p == '\n') {
+      *p = '\0';
+      const char *names[] = {"type", "data"};
+      SEXP elt = PROTECT(Rf_mkNamed(VECSXP, names));
+      SET_VECTOR_ELT(elt, 0, Rf_mkString(con->description));
+      SET_VECTOR_ELT(elt, 1, Rf_mkChar(data->line));
+
+      SEXP vec = VECTOR_ELT(data->output, 2);
+      int len = Rf_length(vec);
+      vec = PROTECT(Rf_lengthgets(vec, len + 1));
```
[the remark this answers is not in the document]
georgestagg: Great idea!

I've set it to grow exponentially, similar to glue's `set` function (https://github.com/tidyverse/glue/blob/main/src/glue.c#L7).

packages/webr/src/outputconnection.c
```diff
@@ -0,0 +1,101 @@
...
+  con->vfprintf = &output_vfprintf;
+  con->destroy = &output_destroy;
+
+  struct output_con_data* data = malloc(sizeof (struct output_con_data));
+  data->output = out;
+  data->cur = data->line = data->buf;
+  con->private = data;
+}
+
+SEXP new_output_connections() {
```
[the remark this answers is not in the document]
georgestagg: I've made an attempt to match this style with `decl/outputconnection-decl.h` and reordering `outputconnection.c`. Let me know if any further tweaks to the organisation of the function declarations are required to match the rlang/vcts style.

src/webR/webr-main.ts
georgestagg: For the moment I've created a function in `utils.ts` named `unpackScalarArrays` that does this job more generally, but I've not fully convinced myself it's the right solution going forward.

I recognise that really I'm trying to work around the fact that scalar elements in R are vectors of length 1. I think the correct solution going forward is a separate PR implementing an option that can be passed to `RObject`'s `toJs` or `toArray` functions (perhaps just for atomic vectors? You probably always want a list to convert into an Array) that changes the conversion behaviour so that R vectors of length 1 are converted into a single JS object, rather than a JS Array of one. There's type instability introduced there, but at least it's behind an opt-in.

src/webR/utils.ts
```diff
@@ -20,3 +22,20 @@ export function promiseHandles() {
 export function sleep(ms: number) {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
+
+export function unpackScalarArrays(obj: RawType): RawType {
+  if (typeof obj === 'object' && obj !== null) {
```
lionel-, line 27: Probably should negate this expression and return early. This way the rest of the function can be unindented one level and it doesn't look like `return obj` may be a fallthrough case in the complex branching below.

PR #63, opened 2022-09-13
Reintroduce function call semantics for `RFunction` type objects

georgestagg:
With this PR, as our R proxies are constructed a test is made to see if the R object has the `exec` method. If it does, it is assumed the object is an R function.

Proxies to R functions are then constructed with an empty function as the base target object. With this, the proxy becomes callable, and a proxy `apply` hook can be used.

An `apply` hook is created such that `await foo(...)` simply becomes a shortcut for `await foo.exec(...)`. In future we can add subtle differences in the argument types for calling via the hook vs invoking `exec` directly, if we would like to do so.

The Typescript alias for `RFunction` is updated to include the information that it is callable.

Unit tests are added to check that `RFunction`s are callable, and some other R object is not.

### DIFF
--- a/src/tests/webR/proxy.test.ts
+++ b/src/tests/webR/proxy.test.ts
@@ -1,4 +1,5 @@
 import { WebR } from '../../webR/webr-main';
+import { RDouble, RFunction } from '../../webR/robj';
 import util from 'util';
 
 const webR = new WebR({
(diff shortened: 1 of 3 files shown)


georgestagg:
An interesting thing to consider might be to automatically run `.toJs()` on R objects returned when using the function call syntax, as a further shortcut. The original R proxy could still be obtained using `exec`, if required.

```
> s = await evalRCode('sin')
< Proxy

> await s.exec([0.1, 0.2, 0.3])
< Proxy

> await s([0.1, 0.2, 0.3])
< Float64Array [0.0998..., 0.1986..., 0.2955...]
```

The current version of `exec` and the apply hook already do some magic to convert JS arguments to R objects, so it might make sense for someone just wanting to use R functions in JS.

lionel- (approved):

PR #65, opened 2022-09-20
Rework the "Console" webR app to work with callbacks and manage the async infinite loop

georgestagg:
This PR reworks the included Console app so that it is constructed by providing callbacks for `stdout`, `stderr` etc. This may be useful and easier to use for users more familiar with how Emscripten apps and Pyodide starts up by default (without use of async workers), where callbacks are provided.

 * Default callbacks are included: prompting the user for input with a JS alert, sending output to the JS console, and plotting with `canvas` graphic device. It is expected for users to override these defaults if they use `Console` within their own project.

* The default plotting callback writes stores output in the instance property `canvas`, a `HTMLCanvasElement` that can be appended into the web page. For the moment, plotting throws an error when running under Node, explaining that HTML canvas is not currently available there. There is a node package replicating the Canvas API that could be added later.

* Rather than building a separate output with `esbuild`, `Console` is now included in `webr.mjs` via a re-export.

* The `Console` app manages the async infinite loop around the webR communication channels for the user. The user simply needs to call `Console.run()` to start reading from webR output. It is important this function is not `await`ed, it never returns.

### DIFF
--- a/src/console/console.ts
+++ b/src/console/console.ts
@@ -1,3 +1,4 @@
+import { IN_NODE } from '../webR/compat';
 import { WebR, WebROptions } from '../webR/webr-main';
 
 /**
@@ -45,7 +46,7 @@ import { WebR, WebROptions } from '../webR/webr-main';
  */
 export class Console {
   webR: WebR;
-  canvas: HTMLCanvasElement;
+  canvas: HTMLCanvasElement | undefined;
   #stdout;
   #stderr;
   #prompt;
@@ -80,9 +81,11 @@ export class Console {
     }
   ) {
     this.webR = new WebR(options);
-    this.canvas = document.createElement('canvas');
-    this.canvas.setAttribute('width', '1008');
-    this.canvas.setAttribute('height', '1008');
+    if (!IN_NODE) {
+      this.canvas = document.createElement('canvas');
+      this.canvas.setAttribute('width', '1008');
+      this.canvas.setAttribute('height', '1008');
+    }
     this.#stdout = callbacks.stdout || this.#defaultStdout;
     this.#stderr = callbacks.stderr || this.#defaultStderr;
     this.#prompt = callbacks.prompt || this.#defaultPrompt;
@@ -127,6 +130,9 @@ export class Console {
    * @param {string} exec - The canvas API command as a text string
    */
   #defaultCanvasExec = (exec: string) => {
+    if (IN_NODE) {
+      throw new Error('Plotting with HTML canvas is not yet supported under Node');
+    }
     Function(`this.getContext('2d').${exec}`).bind(this.canvas)();
   };
 
--- a/src/console/console.ts
+++ b/src/console/console.ts
@@ -1,14 +1,162 @@
-import { WebR } from '../webR/webr-main';
+import { WebR, WebROptions } from '../webR/webr-main';
 
-(async () => {
-  const webR = new WebR();
+/**
+ * @callback ConsoleCallback
+ * @param {string} text
+ * @return {void}
+ */
 
-  for (;;) {
-    const output = await webR.read();
-    if (output.type === 'stdout') {
-      console.log(output.data);
-    } else if (output.type === 'stderr') {
-      console.error(output.data);
+/**
+ * Text-based Interactive Console for WebR
+ *
+ * A helper application to assist in creating an interactive R REPL based on
+ * JavaScript callbacks.
+ *
+ * Callback functions ``stdout`` and ``stderr`` are called with a single line
+ * of output as the first argument. The default implementation of `stdout` and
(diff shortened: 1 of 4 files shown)


src/tests/console/console.test.ts
```diff
@@ -0,0 +1,39 @@
...
+  {
+    WEBR_URL: '../dist/',
+  }
+);
+con.run();
+
+test('Start up the REPL and wait for user input prompt', async () => {
+  await con.webR.init();
+  expect(stdout).toHaveBeenCalledWith('Platform: wasm32-unknown-emscripten (32-bit)');
+  expect(prompt).toHaveBeenCalledWith('> ');
```
lionel-, line 22: Interesting way of testing callbacks.

src/tests/console/console.test.ts
```diff
@@ -0,0 +1,39 @@
...
+});
+
+test('Take input and write to stdout', async () => {
+  con.stdin('42');
+  await sleep(500);
```
lionel-, line 27: 😬
Can we have an optional callback passed to `stdin` that would temporarily override `stdOut` for the corresponding output?

Might be a bit tricky because we'd need to coordinate inputs and outputs. We know that for each line of input, we'll be called with at least one prompt (+ if input is incomplete, > if input is complete), but possibly more than one because of recursive prompts like browser. So that makes coordination hard, unless we force R to run in non-interactive mode.

So I guess we'll just sleep 😴

src/webR/webr-main.ts
```diff
@@ -4,6 +4,8 @@ import { BASE_URL, PKG_BASE_URL } from './config';
 import { newRProxy } from './proxy';
 import { RTargetObj, RTargetType, RObject, isRObject } from './robj';
 
+export { Console } from '../console/console';
```
lionel-, line 7: I'm not sure about this but I'll defer to your judgement.

src/tests/console/console.test.ts
lionel-: Even with a callback for specific inputs, we'd still need a timeout in case there is something wrong and our cb isn't called. The practical difference between sleep(500) and timeout(500) is that with the latter we only sleep for 500 when something is wrong, which can speed up the test suite if there are many tests.

lionel- (approved):

src/tests/console/console.test.ts
lionel-: > possibly more than one because of recursive prompts like browser

It should be possible to pass the recursion depth as parameter to webR though? And then we can coordinate.

src/console/console.ts
```diff
@@ -1,14 +1,168 @@
...
+export class Console {
+  webR: WebR;
+  canvas: HTMLCanvasElement | undefined;
+  #stdout;
+  #stderr;
+  #prompt;
+  #canvasExec;
+
+  /**
+   * @constructor
```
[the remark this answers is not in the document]
georgestagg: I'm new to JSDoc/Typedoc myself and still getting to grips with the tags and best practice of exactly what to document and where.

I've tweaked things a little for now. I think it's still verbose but there should be less repetition. In truth any JSDoc strings I've written so far will probably be tweaked further in the future anyone once the doc building infrastructure is in place and I can better see how the output looks in context.

src/console/console.ts
```diff
@@ -1,14 +1,168 @@
...
+
+  /**
+   * Start the webR console
+   *
+   * Start the infinite loop waiting for output from webR and dispatching
+   * callbacks based on the message recieved.
+   *
+   * The promise returned by this asynchronous function never resolves.
+   */
+  async run() {
```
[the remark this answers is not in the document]
georgestagg: Great idea!

src/tests/console/console.test.ts
georgestagg: If the promise resolves it should do so quickly, and if the promise never resolves due to a problem the global jest timeout will kick in to fail the test automatically.

Currently I've set the global jest timeout to be very long (6 mins IIRC) because I was seeing issues where some tests were taking too long in GitHub Actions and timing out, but I believe the timeout can also be set per test so I will come back later and make the timeout longer only on certain tests.

src/tests/console/console.test.ts
```diff
@@ -23,14 +25,16 @@ test('Start up the REPL and wait for user input prompt', async () => {
 });
 
 test('Take input and write to stdout', async () => {
+  waitForPrompt = promiseHandles();
   con.stdin('42');
-  await sleep(500);
+  await waitForPrompt.promise;
```
lionel-, line 30: Nice!

With this setup though, we'll need to be a bit careful with multiline inputs. But this can be figured out later if needed.

lionel- (approved):

PR #67, opened 2022-10-11
Change how R lists, pairlists and atomic vectors are converted to JavaScript by default

georgestagg:
Contains #62.

R lists and pairlists are changed so that the default structure when converting from an R object to a JS object is an array of objects, rather than simply a JS object. For example, `list(a=1,b=2,c=3)` becomes `[{ a: [1] }, { b: [2] }, { c: [3] }]`.

This structure better reflects the flexibility of R lists. Lists containing objects with repeated names or some non-named objects work better in the new default structure, and the order of objects is preserved.

For both types of list, the method `toJs()` now invokes `toArray()` which returns objects in the new structure. What was previously `toArray()` is now named `values()` to match JS's `Object.values`, which has a similar behaviour.

The current behaviour of `toJs()` on lists can be obtained by explicitly invoking the `toObject()` method.

### DIFF
--- a/src/tests/webR/robj.test.ts
+++ b/src/tests/webR/robj.test.ts
@@ -43,8 +43,9 @@ test('Get RObject type as a string', async () => {
 describe('Working with R lists and vectors', () => {
   test('Get R object attributes', async () => {
     const vector = (await webR.evalRCode('c(a=1, b=2, c=3)')).result;
-    const value = await vector.attrs();
-    expect(await value.toJs()).toEqual({ names: ['a', 'b', 'c'] });
+    const value = (await vector.attrs()) as RList;
+    const attrs = await value.toArray();
+    expect(attrs[0]).toEqual({ names: ['a', 'b', 'c'] });
   });
 
   test('Get R object names', async () => {
@@ -68,9 +69,9 @@ describe('Working with R lists and vectors', () => {
   test('Get an item with [ operator', async () => {
     const vector = (await webR.evalRCode('list(a=1+4i, b=2-5i, c=3+6i)')).result;
     let value = await vector.subset('b');
-    expect(await value.toJs()).toEqual({ b: [{ re: 2, im: -5 }] });
+    expect(await value.toJs()).toEqual([{ b: [{ re: 2, im: -5 }] }]);
     value = await vector.subset(3);
-    expect(await value.toJs()).toEqual({ c: [{ re: 3, im: 6 }] });
+    expect(await value.toJs()).toEqual([{ c: [{ re: 3, im: 6 }] }]);
   });
 
   test('Get an item with $ operator', async () => {
@@ -109,7 +110,9 @@ describe('Working with R lists and vectors', () => {
   test('Convert an R pairlist to JS', async () => {
     const result = (await webR.evalRCode('as.pairlist(list(x="a", y="b", z="c"))'))
       .result as RPairlist;
-    expect(await result.toJs()).toEqual({ x: ['a'], y: ['b'], z: ['c'] });
+    expect(await result.toJs()).toEqual([{ x: ['a'] }, { y: ['b'] }, { z: ['c'] }]);
+    expect(await result.toObject()).toEqual({ x: ['a'], y: ['b'], z: ['c'] });
+    expect(await result.values()).toEqual([['a'], ['b'], ['c']]);
   });
 
   test('R list includes method', async () => {
@@ -119,8 +122,9 @@ describe('Working with R lists and vectors', () => {
 
   test('Convert an R list to JS', async () => {
     const result = (await webR.evalRCode('list(x="a", y="b", z="c")')).result as RList;
-    expect(await result.toJs()).toEqual({ x: ['a'], y: ['b'], z: ['c'] });
-    expect(await result.toArray()).toEqual([['a'], ['b'], ['c']]);
+    expect(await result.toJs()).toEqual([{ x: ['a'] }, { y: ['b'] }, { z: ['c'] }]);
+    expect(await result.toObject()).toEqual({ x: ['a'], y: ['b'], z: ['c'] });
+    expect(await result.values()).toEqual([['a'], ['b'], ['c']]);
   });
 
   test('Convert an R double atomic vector to JS', async () => {
--- a/src/tests/webR/robj.test.ts
+++ b/src/tests/webR/robj.test.ts
@@ -12,6 +12,7 @@ import {
   RPairlist,
   RList,
   REnvironment,
+  RCharacter,
 } from '../../webR/robj';
 
 const webR = new WebR({
@@ -44,8 +45,8 @@ describe('Working with R lists and vectors', () => {
   test('Get R object attributes', async () => {
     const vector = (await webR.evalRCode('c(a=1, b=2, c=3)')).result;
     const value = (await vector.attrs()) as RList;
-    const attrs = await value.toArray();
-    expect(attrs[0]).toEqual({ names: ['a', 'b', 'c'] });
+    const attrs = await value.toObject();
+    expect(attrs.names).toEqual(expect.objectContaining({ values: ['a', 'b', 'c'] }));
   });
 
   test('Get R object names', async () => {
@@ -68,16 +69,19 @@ describe('Working with R lists and vectors', () => {
 
   test('Get an item with [ operator', async () => {
     const vector = (await webR.evalRCode('list(a=1+4i, b=2-5i, c=3+6i)')).result;
-    let value = await vector.subset('b');
-    expect(await value.toJs()).toEqual([{ b: [{ re: 2, im: -5 }] }]);
-    value = await vector.subset(3);
-    expect(await value.toJs()).toEqual([{ c: [{ re: 3, im: 6 }] }]);
+    const val1 = (await vector.subset('b')) as RComplex;
+    const obj1 = await val1.toObject();
+    expect(obj1.b).toEqual(expect.objectContaining({ values: [{ re: 2, im: -5 }] }));
+
+    const val2 = (await vector.subset(3)) as RComplex;
+    const obj2 = await val2.toObject();
+    expect(obj2.c).toEqual(expect.objectContaining({ values: [{ re: 3, im: 6 }] }));
   });
 
   test('Get an item with $ operator', async () => {
     const vector = (await webR.evalRCode('list(a="x", b="y", c="z")')).result;
-    const value = await vector.getDollar('b');
-    expect(await value.toJs()).toEqual(['y']);
+    const value = (await vector.getDollar('b')) as RCharacter;
+    expect(await value.toArray()).toEqual(['y']);
   });
 
   test('Get an item using the pluck method', async () => {
@@ -96,7 +100,7 @@ describe('Working with R lists and vectors', () => {
     vector = await vector.set(1, value);
     vector = await vector.set(2, 5);
     vector = await vector.set('c', 6);
-    const result = (await vector.toJs()) as Float64Array;
+    const result = await (vector as RDouble).toTypedArray();
     expect(Array.from(result)).toEqual([4, 5, 6]);
   });
 
@@ -110,9 +114,19 @@ describe('Working with R lists and vectors', () => {
   test('Convert an R pairlist to JS', async () => {
     const result = (await webR.evalRCode('as.pairlist(list(x="a", y="b", z="c"))'))
       .result as RPairlist;
-    expect(await result.toJs()).toEqual([{ x: ['a'] }, { y: ['b'] }, { z: ['c'] }]);
-    expect(await result.toObject()).toEqual({ x: ['a'], y: ['b'], z: ['c'] });
-    expect(await result.values()).toEqual([['a'], ['b'], ['c']]);
+    const arr = await result.toArray();
+    expect(arr[0]).toEqual(expect.objectContaining({ values: ['a'] }));
+    expect(arr[1]).toEqual(expect.objectContaining({ values: ['b'] }));
+    expect(arr[2]).toEqual(expect.objectContaining({ values: ['c'] }));
+    const obj = await result.toObject();
+    expect(obj.x).toEqual(expect.objectContaining({ values: ['a'] }));
+    expect(obj.y).toEqual(expect.objectContaining({ values: ['b'] }));
+    expect(obj.z).toEqual(expect.objectContaining({ values: ['c'] }));
+    const resJs = await result.toJs();
+    expect(resJs.names).toEqual(['x', 'y', 'z']);
+    expect(resJs.values[0]).toEqual(expect.objectContaining({ values: ['a'] }));
+    expect(resJs.values[1]).toEqual(expect.objectContaining({ values: ['b'] }));
+    expect(resJs.values[2]).toEqual(expect.objectContaining({ values: ['c'] }));
   });
 
   test('R list includes method', async () => {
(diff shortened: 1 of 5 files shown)


src/tests/webR/robj.test.ts
```diff
@@ -109,7 +110,9 @@ describe('Working with R lists and vectors', () => {
   test('Convert an R pairlist to JS', async () => {
     const result = (await webR.evalRCode('as.pairlist(list(x="a", y="b", z="c"))'))
       .result as RPairlist;
-    expect(await result.toJs()).toEqual({ x: ['a'], y: ['b'], z: ['c'] });
+    expect(await result.toJs()).toEqual([{ x: ['a'] }, { y: ['b'] }, { z: ['c'] }]);
+    expect(await result.toObject()).toEqual({ x: ['a'], y: ['b'], z: ['c'] });
```
lionel-, line 114: If there are synonyms, it's an error right? Do we have a test for this?

Edit: After seeing the implementation I see that latter synonyms override former ones. As argued below, I think this should be the opposite.

src/webR/robj.ts
```diff
@@ -470,22 +481,27 @@ export class RObjList extends RObjImpl {
...
-      })
-    );
+  values(): RawType[] {
+    return this.toArray().map((obj) => obj[Object.keys(obj)[0]]);
   }
 
-  toJs(): RawType {
-    return this.toObject();
+  toObject(): NamedObject<RawType> {
+    return this.toArray().reduce((prev, cur) => Object.assign(prev, cur), {});
```
lionel-, line 489: oh so in case of synonyms, latter names override former names? This may be a reasonable way of solving this problem but probably former names should have precedence since `$` matches first occurrences rather than last? So maybe reverse the array before reduction?

lionel- (commented):
I like that this better preserves data by default. I assume we could add scanning tools to easily pick the first element matching a name, the way `$` works in R?

This is type stable right? I think it's fine for `toJS()` to use different conversion methods based on the type of an object, but it would seem wrong to do it based on properties of the values (i.e. presence of names). An unnamed list would yield `[{['a']}, {['b']}]`? (This might need a test.) This seems like a lot of wrapping though, maybe too much to be practical.

I still wonder if it wouldn't be better to consistently transform vectors as:

```js
{ names: nil, values: [...]}
{ names: ['a', 'b'], values: [...]}
```

The layout proposed in this PR has the advantage that it's easy to subset both names and values on the JS side, the same way it's easy on the R side using `[`. I'm just not sure it should be the default for `toJS()`, maybe that should be the layout for a new method like `mylist.zip()`?

It would make sense to add `zip()` methods to atomic vectors too. With the understanding that this is an inefficient operation in the atomic case since it disaggregates the data (for names the disaggregation is fine since character vectors are stored as lists of interned strings to begin with).

So I think my proposal is:

- Add `toArrays()` methods to all vectors including lists. This would return an object with `names` and `values` fields. This is a convenient way of getting both `names()` and `values()`.

- Add `zip()` methods to all vectors including lists, meaning zip the names and values into an array of objects.

- Use `toArrays()` as the default `toJS()` method for lists (and perhaps all vectors). This way you can easily see all your data when you're working interactively with R objects in the JS console. This representation has the advantage of being easy to use even when there are no names.

lionel-:
For atomic vectors, maybe `toArrays()` could also include a boolean array of missingness indicators. The idea would be to produce an overview of the data in an object for interactive usage and draw attention to potential edge cases, such as presence of missing values or absence of names.

src/webR/robj.ts
georgestagg: Yep, good spot!

lionel-: I wonder if we should add an optional argument to be stricter and throw an error if there are duplicate names? This would allow programming more defensively.

georgestagg:
I do like the look of `{ names: ['a', 'b'], values: [...]}`. You had mentioned this structure before, but I had forgotten about it! I like how we can add on extra properties later, such as the indicators you suggest. I will implement it and see how it feels.

I might also rework the originally proposed structure into something that looks like what you get from JS's `Object.entries()`, returning `[[ 'a', [1] ], [ 'b', [2] ]]`. I didn't like that structure for the default `toJs()`, but it might be a good version for a non-default as `zip()` or `entries()`, as you could then use it directly with `Object.fromEntries()` on the JS side.

src/webR/robj.ts
georgestagg: Probably a good idea, yes. At some point I need to consider a general `options` argument for `toJs()` and friends, so that things like that can be propagated through the recursive conversion process.

src/webR/utils.ts
```diff
@@ -23,18 +23,31 @@ export function sleep(ms: number) {
...
+  const names = obj.names as string[];
+  if (!names || names.some((name) => typeof name === 'undefined')) {
+    throw new Error('Attempted to merge unnamed list array');
+  }
+  return obj.values
+    .map((v, idx) => {
+      return { [names[idx]]: v };
+    })
+    .reduce((prev, cur) => Object.assign(prev, cur), {});
+}
```
georgestagg, line 53: Note that this function, `mergeListArray`, goes away in #73. With that PR it's no longer required in `webr-main.ts`, the newly enabled `for await ... of` syntax is used instead.

src/webR/robj.ts
```diff
@@ -565,8 +617,10 @@ type TypedArray =
   | Int32Array
   | Uint32Array
   | Float32Array
-  | Float64Array
-  | Array<Complex>;
+  | Float64Array;
+
+type Logical = boolean | 'NA' | undefined;
```
lionel-, line 622: When do we use `undefined` here? Could this be `boolean | null`?

src/webR/robj.ts
```diff
@@ -585,27 +639,62 @@ abstract class RObjAtomicVector extends RObjImpl {
     throw new Error('$ operator is invalid for atomic vectors');
   }
 
-  abstract toArray(): TypedArray;
+  missing(): boolean[] {
```
lionel-, line 642: In vctrs we developed the convention that `is_foo()` is for scalar predicates and `detect_foo()` is for vectorised predicates. So if we want to use a verb here we could do `detectNa()` or `detectMissing()`. This is longer but might better communicate semantics if we use this naming scheme more generally.

src/webR/robj.ts
```diff
@@ -585,27 +639,62 @@ abstract class RObjAtomicVector extends RObjImpl {
...
+    const isna = Module.allocateUTF8('is.na');
+    const call = Module._Rf_protect(Module._Rf_lang2(Module._Rf_install(isna), this.ptr));
+    const val = RObjImpl.wrap(Module._Rf_eval(call, RObjImpl.baseEnv.ptr)) as RObjLogical;
+    val.preserve();
+    const ret = val.toTypedArray();
```
lionel-, line 647: Can `toTypedArray()` throw? Or the computation be interrupted in any way? If so we'll leak `val`. This is a general problem so no need to fix here, but it would be nice to find an unwind strategy for cleaning up consistently this sort of code in case of unexpected jumps.

By the way why use `protect()` for `call` and `preserve()` for `val`? I'd only use preserve if the object escapes the current context or is preserved globally.

Note that the issue of cleaning up concerns both preserve and protect.

src/webR/robj.ts
```diff
@@ -585,27 +639,62 @@ abstract class RObjAtomicVector extends RObjImpl {
...
+    Module._free(isna);
+    return Array.from(ret).map((elt) => Boolean(elt));
+  }
 
-  toObject(): { [key: string | number]: RawType } {
-    const names = this.names();
+  abstract toTypedArray(): TypedArray;
+
+  toArray(): atomicType[] {
+    return Array.from(this.toTypedArray());
```
lionel-, line 657: Deal with `NA` here?

I think our strategy for missing values could be:

- We leave data intact in typed arrays, this is meant for efficient usage. Must be paired with the missingness predicate.
- We convert missing values to `null` when creating classic arrays.

lionel- (commented):
Looks great!

I still have a couple of concerns about the conversion.

- In `toObject()`, using indices for unnamed elements causes ambiguity with names. I wonder if we should just require fully named vectors in that conversion method.

- I think we should use `null` instead of undefined for missing values when converting to classic arrays.

- `""` names should convert to `''`

src/webR/robj.ts
georgestagg: I have not been super consistent with my choice of `protect` vs `preserve` generally. Going forward I'll try and stick to `protect` unless the object is to live outside the current function.

src/tests/webR/robj.test.ts
```diff
@@ -218,7 +310,10 @@ describe('Working with R environments', () => {
   test('Convert an R environment to JS', async () => {
     const env = (await webR.evalRCode('x<-new.env();x$a=TRUE;x$b=FALSE;x$.c=NA;x'))
       .result as REnvironment;
-    expect(await env.toJs()).toEqual({ '.c': ['NA'], a: [true], b: [false] });
+    const envJs = await env.toJs();
+    expect(envJs['.c']).toEqual(expect.objectContaining({ values: ['NA'] }));
```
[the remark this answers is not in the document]
lionel-: ah no I guess that `list(TRUE, NA, FALSE)` would convert to `[[true], [null], [false]]`? So the ambiguity is between `list(list(NULL))` and `list(NA)` which would both convert to `[[null]]`.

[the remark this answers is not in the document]
georgestagg: I'm currently changing this to avoid turning integer NA's into 'NA' string. Atomic vectors will be more consistent, with converted atomic types becoming `T | null`, where `T` is e.g. `boolean` in the case of a logical vector. `toArray()` will return a normal array containing converted values of this type.

In my current build I have it returning the `value` property as typed arrays with R's untouched sentinel values by default. I wanted to avoid the overhead of conversion, but I think it makes sense for complex and string vectors to override this and return converted values instead since the raw typed array does not make much sense without knowledge of R internals.

I am still unsure, but I could be convinced that `toArrays()` should always returns converted values in the `values` property if it makes more sense, bearing in mind there will be some conversion overhead. One can always use `toTypedArray()` to get to the raw buffer.

In my current build, `list(TRUE, NA, FALSE)` becomes,

```
{
  names: undefined,
  values: [
    { values: Int32Array(1), missing: [false], names: undefined },
    { values: Int32Array(2147483648), missing: [true], names: undefined },
    { values: Int32Array(0), missing: [false], names: undefined }
  ]
}
```

So, `list(NA)` becomes,
```
{
  names: undefined,
  values: [
    { values: Int32Array(2147483648), missing: [true], names: undefined },
  ],
}
```

and `list(list(NULL))` becomes,
```
{
  names: undefined,
  values: [
    {values: [null], names: undefined}
  ]
}
```

Later I would like the result of `toArrays()` to be able to be converted back into an R object, and survive a round trip if possible. Just now I was worried about `list(as.character(NA))` being ambiguous, but that gives the slightly different,
```
{
  names: undefined,
  values: [
    { values: [null], missing: [true], names: undefined }
  ]
}
```
with the missing array differentiating between `NULL` and `NA`.

One option to avoid ambiguity for sure, and perhaps to make things easier when we want to convert back, is to add a `type` property to the result of `toArrays()`, with either the `RType` enum number or a string. I think that is probably worth doing.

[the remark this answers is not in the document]
lionel-: Ah gotcha. This makes `toArrays()` having a flavour of being a method for interop rather than for interactivity. Another possibility is to have both `toArrays()` and `toTypedArrays()`. The former would offer a more convenient and compact representation for discovery, development, and debugging, but would have some overhead.

I like the idea of adding type information to remove ambiguity.

So with `toTypedArrays()`, `list(TRUE, NA, FALSE)` would look mostly like your current output, with type info added. With `toArrays()` we'd get:

```r
{
  type: 'list',
  names: undefined,
  values: [
    { type: 'logical', values: [true], names: undefined },
    { type: 'logical', values: [null], names: undefined },
    { type: 'logical', values: [false], names: undefined }
  ]
}
```

It's more important to first have adequate representation, so `toArrays()` could come later and we could just export `toTypedArrays()` to begin with.

By the way, using `null` instead of `undefined` for names seems more consistent with the R representation. I was thinking that when in doubt between a representation closer to JS and one closer to R, opting for the latter as a rule would make things more consistent and predictable.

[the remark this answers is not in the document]
lionel-: Two more thoughts:

- The choice between classic JS arrays with `null` missing values and typed arrays with an extra array of missingness indicators could be made with an option instead of having two different methods.

- I'm not sure `toArrays()` still makes sense as a name. Really we're converging towards a full representation of R objects. I expect we'll add `attributes` field for instance. Maybe it should be called something like `serialise()`?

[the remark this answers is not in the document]
lionel-: One more:

* For deeply nested objects, it will be useful to have a `depth` option to control the depth of recursion.

src/tests/webR/robj.test.ts
```diff
@@ -119,35 +136,110 @@ describe('Working with R lists and vectors', () => {
 
   test('Convert an R list to JS', async () => {
     const result = (await webR.evalRCode('list(x="a", y="b", z="c")')).result as RList;
-    expect(await result.toJs()).toEqual({ x: ['a'], y: ['b'], z: ['c'] });
-    expect(await result.toArray()).toEqual([['a'], ['b'], ['c']]);
+    const arr = await result.toArray();
+    expect(arr[0]).toEqual(expect.objectContaining({ values: ['a'] }));
```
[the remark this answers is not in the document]
georgestagg: I've tried to access `values` directly on objects as much as possible, but when grabbing nested objects (i.e. an object at least 1 level deep inside a list) the type system can't know the type of the R object until runtime. When that happens, the returned object could be `null` from an R `NULL`, and accessing the `values` property of that is an error.

There's a few workarounds, e.g. checking if `'values'` exists first and failing the test if not, setting the object type explicitly with `as` (since here we know what the R object looks like), etc. I have instead used Jest's `objectContaining` as a workaround.

I've removed some of the use of `objectContaining` in the latest commit, but where I have kept it I've added `names: null` as part of the test.

src/tests/webR/robj.test.ts
```diff
@@ -119,35 +136,110 @@ describe('Working with R lists and vectors', () => {
...
+  test('Partially undefined names attribute', async () => {
+    const list = (await webR.evalRCode('list(x="a", y="b", "c")')).result as RList;
+    const pairlist = (await webR.evalRCode('pairlist(x="a", y="b", "c")')).result as RPairlist;
+    const atomic = (await webR.evalRCode('c(x="a", y="b", "c")')).result as RCharacter;
+    const listJs = await list.toArrays();
+    const pairlistJs = await pairlist.toArrays();
+    const atomicJs = await atomic.toArrays();
+    expect(listJs.names).toEqual(['x', 'y', undefined]);
+    expect(pairlistJs.names).toEqual(['x', 'y', undefined]);
+    expect(atomicJs.names).toEqual(['x', 'y', undefined]);
```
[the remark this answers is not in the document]
georgestagg: Empty names are now left as `''`, this test now checks for the names array to be `['x', 'y', '']`.

In character vectors `NA` now becomes `null` when converted, as with all other atomic vectors.

A test has been added for NA names. In this case the names array becomes something like `['foo', ..., null])`.

src/tests/webR/robj.test.ts
```diff
@@ -119,35 +136,110 @@ describe('Working with R lists and vectors', () => {
...
+  });
+
+  test('Automatic indexing when converting to JS object', async () => {
```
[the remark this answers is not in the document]
georgestagg: For the moment, I've removed the automatic indexing in favour of encouraging fully named vectors for `toObject` conversion. Your example has convinced me that the automatic indexing can not reasonably work unless we can independently set `foo[2]` and `foo['2']` with a JS object.

There is a new option `allowEmptyKey` in `toObject`, defaulting to `false`. If the user _really_ wants to, it can be set to `true`. The resulting object is allowed a key `''` or `null` (depending on if the name in R was empty or `NA`), and the first value in the list with that name wins.

PR #68, opened 2022-10-11
Switch from jquery terminal to xterm.js with xterm-readline

georgestagg:
Includes #62.

This should provide improved performance. I'll do some basic measurements and add them here soon.

EDIT:
Time taken to print n elements of random `rnorm` output at 80 characters per line:
| n | jquery.terminal (ms) | xterm.js (ms)
|---|---|---|
| 100   | 24 | 12 |
| 1000   | 92 | 143 |
| 10000  | 1932 | 350 |
| 100000  | 318771 | 2171 |
| 1000000  | ? | 20963 |

### DIFF
--- a/src/package.json
+++ b/src/package.json
@@ -5,9 +5,11 @@
   "dependencies": {
     "@types/emscripten": "^1.39.6",
     "jquery": "^3.6.0",
-    "jquery.terminal": "^2.33.3",
     "jstree": "^3.3.12",
-    "xmlhttprequest": "git+https://github.com/georgestagg/node-XMLHttpRequest.git#response-buffer"
+    "xmlhttprequest": "git+https://github.com/georgestagg/node-XMLHttpRequest.git#response-buffer",
+    "xterm": "^4.19.0",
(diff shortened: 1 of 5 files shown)


src/package.json
```diff
@@ -5,9 +5,11 @@
   "dependencies": {
     "@types/emscripten": "^1.39.6",
     "jquery": "^3.6.0",
```
wch, line 7: You can probably remove `jquery` here.

src/tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "noUnusedLocals": true,
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
-    "types": ["node", "emscripten", "jquery", "jquery.terminal", "jstree", "jest"],
+    "types": ["node", "emscripten", "jquery", "xterm", "jstree", "jest"],
```
wch, line 12: I think that you shouldn't need `jquery` or `xterm`.

lionel- (approved):
Nice speed up. It's great seeing how easy it is to change the terminal front!

PR #76, opened 2022-10-20
Enable interruption of long running R computations

georgestagg:
R will periodically call `R_PolledEvents` during long running computations. A handler is installed there to manage events from webR.

A `SyncTask` of type `userBreakSignal` is setup to check for interrupts as part of the event handling. If the user has signalled an interrupt by invoking `webR.interrupt()` (also hooked up to `Ctrl-C` in the REPL app), R's `onintr()` is invoked which breaks execution and returns to the prompt.

### DIFF
--- a/src/webR/chan/channel.ts
+++ b/src/webR/chan/channel.ts
@@ -35,7 +35,7 @@ if (IN_NODE) {
 export class ChannelMain {
   inputQueue = new AsyncQueue<Message>();
   outputQueue = new AsyncQueue<Message>();
-  userBreakSignal = 0;
+  #interruptBuffer?: Int32Array;
 
   initialised: Promise<unknown>;
   resolve: (_?: unknown) => void;
@@ -96,6 +96,13 @@ export class ChannelMain {
     return msg;
   }
 
+  interrupt() {
+    if (!this.#interruptBuffer) {
+      throw new Error('Failed attempt to interrupt before initialising interruptBuffer');
+    }
+    this.#interruptBuffer[0] = 1;
+  }
+
   write(msg: Message) {
     this.inputQueue.put(msg);
   }
@@ -140,6 +147,7 @@ export class ChannelMain {
 
     switch (message.type) {
       case 'resolve':
+        this.#interruptBuffer = new Int32Array(message.data as SharedArrayBuffer);
         this.resolve();
         return;
 
@@ -163,11 +171,6 @@ export class ChannelMain {
             await syncResponse(worker, reqData, response);
             break;
           }
-          case 'userBreakSignal': {
-            await syncResponse(worker, reqData, this.userBreakSignal);
-            this.userBreakSignal = 0;
-            break;
-          }
           default:
             throw new TypeError(`Unsupported request type '${payload.type}'.`);
         }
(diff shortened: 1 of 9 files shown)


src/webR/chan/channel.ts
```diff
@@ -35,6 +35,7 @@ if (IN_NODE) {
 export class ChannelMain {
   inputQueue = new AsyncQueue<Message>();
   outputQueue = new AsyncQueue<Message>();
+  userBreakSignal = 0;
```
lionel-, line 38: I think we can probably refer to this as `interrupt`.

The term `userBreak` only appears in the R platform implementation of Windows.

src/webR/chan/channel.ts
```diff
@@ -197,4 +205,9 @@ export class ChannelWorker {
     const task = new SyncTask(this.#ep, msg);
     return task.syncify() as Message;
   }
+
+  userBreakSignal(): number {
+    const task = new SyncTask(this.#ep, { type: 'userBreakSignal' });
+    return task.syncify() as number;
+  }
```
lionel-, line 212: If we change to "interrupt", this would change too.

lionel- (approved):
IIUC we have the following setup:

- New API point `interrupt()` on main. When called, sets a flag in the main side of the channel.
- We hook into PolledEvents on the worker side.
- When polled, we send a request to the main side to ask if we need to interrupt.

The communication works this way, rather than the other way around, because we currently have no way of modifying the worker state while it is busy.

Another approach would be to set a SharedArrayBuffer with a boolean flag that main could set and worker inspect without any channel communication. This way we avoid unnecessary messages during polled events. Instead of a boolean flag, this could also be an integer represent an enum of events, for extensibility.

This implementation is fine for now though, but maybe add some notes about this?

georgestagg:
> Another approach would be to set a `SharedArrayBuffer` with a boolean flag that main could
> set and worker inspect without any channel communication. This way we avoid unnecessary
> messages during polled events.

I did think about doing something like this, but later I think we're going to want to expand this to have an entire event queue with events containing data more complex than just a boolean or integer.

How about a SharedArrayBuffer set up in this way on the worker to simply act as a signal that there are events waiting in an event queue that lives on the main thread? The main thread sets the flag when there is an event added and the worker clears it once it's emptied the queue. That way the worker can poll the local signal first and only initiate a full message when there is something to consume. With this we can add events that require a more complex data transfer later.

lionel-:
Good idea.

georgestagg:
I've switched things around to use `SharedArrayBuffer` to signal interrupt. The shared buffer that was already setup in `task-worker.ts` is sent to the main thread instead, and a function that calls `Module._Rf_onintr()` is passed to the `setInterruptHandler` function also there. Functionality should be the same, without the heavier message passing.

src/webR/chan/channel.ts
```diff
@@ -155,14 +164,16 @@ export class ChannelMain {
...
-          throw new TypeError("Unsupported request type '$(payload.type)'.");
+        switch (payload.type) {
+          case 'read': {
+            const response = await this.inputQueue.get();
+            // TODO: Pass a `replacer` function
+            await syncResponse(worker, reqData, response);
+            break;
+          }
+          default:
+            throw new TypeError(`Unsupported request type '${payload.type}'.`);
```
georgestagg, line 175: This change is now functionally redundant, but I'm going to leave it in place.

I like the idea of a lightweight shared buffer signal leading to a heavier message passing system for passing events to the worker thread in the future, and when that's implemented this will need to be changed to a `switch` statement in any case.

PR #79, opened 2022-10-26
Add depth parameter to `toTree`, `toArray`, `toObject` on R lists, pairlists and environments

georgestagg:
The depth parameter controls how deep R objects are converted to JS Objects. 0 means infinite depth. The argument default is set to 0 for `toTree` and 1 for `toArray` and `toObject`.

Getting this right was surprisingly tricky, as the previous implementation of `newRProxy`,the `RProxy` type, and the `RProxify` mapped type were not set up for converting nested instances of `RObjImpl` into `RObject` during transfer from the worker thread to the main thread. With this PR the returned objects are recursively searched for pointers to R objects (using the new `isRTargetPtr`), with each found object wrapped in a `RProxy` rather than just checking the outermost object.

Searching through objects and replacing things that match a pattern now occurs in a few different places, so I have abstracted it into a `replaceInObject` utility function that tests for objects matching a filter and applies a replacer function if it finds them.

On the worker thread, in addition to updating methods to take a `depth` argument, the `RObjectTree` type is reorganised to take into account that `toTree()` could return items in the `values` property that are any one of:

 * `null`
 * A JS object of some kind, i.e type `RawType`
 * Unconverted R objects, i.e type `RObjImpl`
 * Tree objects one level deeper, i.e. `RObjectTree`

The`RObjectTree<RObjImpl>` type is used for tree objects on the worker thread and can contain instances of `RObjImpl`. `RObjectTree<RObject>` used for tree objects on the main thread and can contain instances of proxied `RObject` instead. The new `RObjectTreeAtomic` type is used for atomic vectors where objects cannot be nested.

`RObjectTree<RObjImpl>` types are automatically converted to `RObjectTree<RObject>` types when they are passed to the main thread by adding a test in the `RProxy` mapped type in `proxy.ts`.

Finally, unit tests are added to check that `{ depth: 1 }` behaves as expected. Older unit tests relying on the old behaviour to check nested values inside returned objects have `{ depth: 0 }` added where required.

### DIFF
--- a/src/webR/robj.ts
+++ b/src/webR/robj.ts
@@ -882,6 +882,23 @@ export function isRObject(value: any): value is RObject {
 }
 
 /**
+ * Test for an RTargetPtr instance
+ *
+ * @param {any} value The object to test.
+ * @return {boolean} True if the object is an instance of an RTargetPtr.
+ */
+export function isRTargetPtr(value: any): value is RTargetPtr {
+  return (
+    value &&
+    typeof value === 'object' &&
+    'type' in value &&
+    'obj' in value &&
+    'methods' in value &&
+    value.type === 'PTR'
+  );
+}
+
+/**
  * Test for an RFunction instance
  *
  * @param {any} value The object to test.
--- a/src/webR/robj.ts
+++ b/src/webR/robj.ts
@@ -88,6 +88,33 @@ type Complex = {
   im: number;
 };
 
+export type RawType =
+  | number
+  | string
+  | boolean
+  | undefined
+  | null
+  | Complex
+  | Error
+  | ArrayBuffer
+  | ArrayBufferView
+  | Array<RawType>
+  | Map<RawType, RawType>
+  | Set<RawType>
+  | { [key: string]: RawType };
+
+export type NamedEntries<T> = [string | null, T][];
+export type NamedObject<T> = { [key: string]: T };
+
+export type RObjectTree<T> = RObjectTreeImpl<(RObjectTree<T> | RawType | T)[]>;
+export type RObjectTreeAtomic<T> = RObjectTreeImpl<(T | null)[]>;
+type RObjectTreeImpl<T> = {
+  type: string;
+  names: (string | null)[] | null;
+  values: T;
+  missing?: boolean[];
+};
+
 function newRObjFromTarget(target: RTargetObj): RObjImpl {
   const obj = target.obj;
   if (target.type === RTargetType.PTR) {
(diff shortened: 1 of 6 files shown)


src/webR/robj.ts
```diff
@@ -433,26 +445,19 @@ export class RObjSymbol extends RObjImpl {
...
-  toArray(): RawType[] {
-    return this.toTree().values;
+  toArray(options: ToTreeOptions = { depth: 1 }): (RObjImpl | RawType | RObjectTree<RObjImpl>)[] {
+    return this.toTree(options).values;
   }
 
-  toObject({ allowDuplicateKey = true, allowEmptyKey = false } = {}): NamedObject<RawType> {
-    const entries = this.entries();
+  toObject({ allowDuplicateKey = true, allowEmptyKey = false, depth = 1 } = {}): NamedObject<
+    RObjImpl | RawType | RObjectTree<RObjImpl>
```
lionel-, line 458: It seems worth it to alias that triplet union to a single name, since it's used in so many places.

src/webR/robj.ts
```diff
@@ -891,6 +940,25 @@ export function isRFunction(value: any): value is RFunction {
   return isRObject(value) && 'methods' in value._target && value._target.methods.includes('exec');
 }
 
+/**
+ * Test for an RObjectTree instance
+ *
+ * @param {any} value The object to test.
+ * @return {boolean} True if the object is an instance of an RObjectTree.
+ */
+export function isRObjectTree(value: any): value is RObjectTree<any> {
```
lionel-, line 949: Thought: If we control the creation of these objects, would it be better to add a property containing a type label that'd be easier to test for? IIUC, the only advantage of checking the structure of objects instead of a label is to provide duck-typing behaviour and allow flexible creation of these objects by external packages and users.

lionel- (approved):

src/webR/robj.ts
georgestagg: The main use I have for these functions is for TypeScript narrowing using [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates), and I wrote them with an idea of trying to be as robust as possible, since the input argument has been set as type of `any` and my TypeScript linter is a little strict about working with `any` types.

You're right in that this and the other similar functions can probably be simplified a little without causing any issues. In the case of `RObjectTree`, they have a `type` property that should be one of the `RType`s, and checking for that is probably sufficient. I'll take a look a tidying them up before merging.

lionel-: I was thinking about namespaced properties like `webRObjectTree: true`. I.e. a property that unambiguously signals an object created by one of our constructors so that we can assume that its structure and typing is correct.

Otherwise, I think a full check of the object structure (as implemented now) is probably better? Since that implies anyone can construct these objects.

PR #80, opened 2022-11-01
Make ChannelMain and ChannelWorker abstract communication channel interfaces

georgestagg:
This PR moves the current `channelMain` and `channelWorker` classes based on `SharedArrayBuffer` into `SharedBufferChannelMain` and `SharedBufferChannelWorker`.

`ChannelMain` and `ChannelWorker` are now interfaces which other channel schemes can implement to be used as alternatives when `SharedArrayBuffer` is unavailable.

Functions `newChannelMain` and `newChannelWorker` are used to select which class implementing the channel interface is actually used. In this PR `SharedBufferChannel` is used if it is possible, otherwise an error is thrown.

The Emscripten `Module` object now has `noInitialRun` set `true`, and Emscripten's `callMain` function is exposed. This allows us to choose if and when we want to run R's `main()` entrypoint. It is not so useful in this PR, but later becomes useful for fallback communication channels where we might not want to run the normal R REPL because we cannot block there. This seems preferable to previous tricks I have tried, such as calling `exit()` to stop computation.

This PR sets things up so that it is up to the channel implementation to startup R, using `main()` or otherwise. (See `src/unix/Rembedded.c` in R source for an example of how to run R code _without_ running R's `main()`).

The previous `inputOrDispatch` function is tweaked so that while the dispatch code is still implemented in `webr-worker.ts`,  but passed as a callback to the channel implementation. This allows the channel to decide how to handle input and dispatch messages. For `SharedBufferChannel` the same method is the same as before (looping and dispatching over the input queue until a `stdin` message is found), but later if we have a channel that cannot block inside the REPL we'll need to dispatch messages at a different point in the main worker loop.

A few other minor changes:
 * `task-worker.ts` is tweaked so that it does not try to use `SharedArrayBuffer` immediately at import time. Required when we're using other channel types later.
 * The cross-origin URL worker loading trick is moved to `utils.ts` so it can be used by multiple channels without code duplication.


src/webR/chan/channel-sharedbuffer.ts
```diff
@@ -0,0 +1,211 @@
+import { AsyncQueue } from './queue';
```
lionel-, line 1: I'm assuming most of the changes here concern names?
It's sad that we're losing the VCS history even though the code is the same. To avoid this:

- First rename the file without any changes. This can be pushed as a separate branch that doesn't require reviewing.
- Then update the new file with the name changes.

lionel- (approved):

src/webR/chan/channel-sharedbuffer.ts
georgestagg: I've made a new branch `rename-channel` that just renames `channel.ts` to `channel-sharedbuffer.ts` (and tweaks the imports so that it still builds), then changed the base of this PR to that branch. Hopefully that preserves the history better and I can just merge them in turn?

lionel-: Looks like this did the trick!

PR #82, opened 2022-11-01
Communication channel based on postMessage

georgestagg:
Includes #81.

A fallback communication channel is added based on `postMessage`. It is automatically used when both `SharedArrayBuffer` and service workers are unavailable (e.g. loading webR from CDN without the HTTP cross-origin isolation headers set).

With this scheme we cannot block in the REPL, so a warning is issued to the JS console at startup.

On startup, rather than starting R's REPL as normal with `main()`, R is initialised as it is in `src/unix/Rembedded.c` from the R source, using the `Rf_initialize_R()`, `setup_Rmainloop()`, and `R_ReplDLLinit()` functions. CLI arguments are manually allocated in WASM memory, since we are not using Emscripten's `callMain()`.

R's existing `R_ReplDLLdo1` function is used to prompt the user for input. The channel is setup to send a message to the main thread to prompt the user for input, but then return without waiting. By returning with a value of -1 at this point, rather than blocking, we can use this function to create a fallback asynchronous REPL using an infinite loop inside an async Javascript function. 

After prompting, the fallback REPL next sends a request to the main thread asking for user input and `await`s until there is a response. Since we are `await`ing here, rather than blocking inside C code, we yield to the event loop and can receive messages back from the main thread via `postMessage`.

Once some input has arrived from the main thread, it is copied to WASM memory and then executed, again using R's existing `R_ReplDLLdo1` function. Attempts to jump back to the top level during an R error are caught handled by restarting the REPL with `R_ReplDLLinit`.

---

This scheme has some drawbacks:
 * Nested REPLs (`menu`, `browser`, `readline`, etc.) do not work. Things are setup so a descriptive
    error is thrown if you try.
 * Long running R computations cannot be interrupted.
 * Possibly others?

As a consequence of the drawbacks, the webR REPL app is updated so that the global prompt that is shown when one tries to load a package that has not yet been installed with `webr::install()` does not attempt to show an R `menu()` when webR is loaded in postMessage mode.

Finally, minimal unit tests are added. They are duplicates of currently existing tests that simply check that the fallback `postMessage` based channel is able to communicate by sending input and receiving output from the worker thread.


lionel- (commented):
This is only a POC right?

Since this approach has considerable downsides (more patches to R, added complexity, limitations of functionality) I was thinking we'd only experiment with this after getting feedback with the service worker approach.

georgestagg:
Yes, I pushed ahead with this implementation as a way to ensure that (if we decided to go for something like this) it would fit well into the abstracted channel interfaces designed the earlier PRs, and as way to tangibly measure what the benefits vs. drawbacks would be.

It would require further work for merging and we'd be looking at adding it sometime after the first versioned release. On reflection, it probably should have remained an experimental branch rather than a PR.

As an update, today I think I figured a reasonably nice way to get the service worker method working cross-origin and without having to set HTTP headers. If it works, we'll have blocking in the worker thread in all situations except for loading from `file://` or in the case of an existing service worker. With cross-origin service workers I no longer think think the benefits of this PR justify the drawbacks, and I will close (but not delete the branch) unless we get further feedback requesting it after release.

PR #83, opened 2022-11-02
WIP: Expand support for creating R objects from the main thread

georgestagg:
Includes #79.

This PR expands support for creating R objects from the main thread using `webR.newRObject()`. Creating lists, pairlists, atomic vectors, and environments is supported.

Atomic vectors can be created from JS scalars or arrays containing identical types. Lists and pairlists can be created from JS arrays with mixed types. Environments can be created from JS objects using the key/value pairs.

Creating new R objects from JS objects passed to `webR.newRObject()` in the form of an `RObjectTree<T>` (i.e. the type of the result obtained by running `RObjImpl.toTree()`) is also supported, and so round trip conversions are possible.

`RProxy` objects can also be passed to `webR.newRObject()` to be included in lists or environments.

The `RObjImpl` subclass constructors each create an R object of that type from a compatible JS object or `RTargetObj ` (derived from an `RProxy` passed from the main thread). Creating an object using the general `RObjImpl` constructor instead invokes `newRObjFromTarget` to try and choose the best type of object to create, and then invokes that type's subclass constructor.

A few other utility changes:
 * `RObjImpl.type()` is now a method so that it can be accessed from the main thread.
 * `names()` now takes an optional argument to set the names attribute.
 * `REnvironment` now has a `define(name, value)` method to add an object to an environment. *(Note: this was originally used in `newRObjFromTarget`, but no longer used there. However, it seems a shame to remove the method again now.)*
 * `RObjImpl.naInt`/`naReal`/`naString` returns the `NA` sentinels.

TODO:
 * Passing in a list of lists should be supported for creating objects using key/value pairs.

### DIFF
--- a/src/tests/webR/robj.test.ts
+++ b/src/tests/webR/robj.test.ts
@@ -60,6 +60,13 @@ describe('Working with R lists and vectors', () => {
     expect(await value).toEqual(null);
   });
 
+  test('Set R object names', async () => {
+    const vector = (await webR.evalRCode('c(1, 2, 3)')).result;
+    await vector.names(['d', 'e', 'f']);
+    const value = await vector.names();
+    expect(await value).toEqual(['d', 'e', 'f']);
+  });
+
   test('Get an item with [[ operator', async () => {
     const vector = (await webR.evalRCode('list(a=1, b=2, c=3)')).result;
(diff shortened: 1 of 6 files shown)


georgestagg:
Closed in favour of splitting into several independent PRs.

PR #85, opened 2022-11-18
Install parallel package

lionel-:
Needed for tinytest, and in turn to test packages like digest.

Hopefully this doesn't break anything? I couldn't see any negative effect after applying this patch.


georgestagg (approved):
Happy to re-include `parallel` in the default packages, I don't think keeping it breaks anything. However it doesn't completely work, either. For example, `parallel::mcparallel(1:10)` errors out due to lack of `fork` under Emscripten.

This means we can't add the `parallel` package to the base package tests in `src/tests/packages/webr.test.ts` unless we patch out some tests as part of `patches/R-4.1.3/emscripten-avoid-testing-issues.diff`. Though if we did that there's probably nothing left to test anyway.

Probably the right thing to do is to make an Issue to track this and modify the package later to avoid the `fork()` call on Emscripten.

PR #86, opened 2022-11-18
Rework `enum`s into objects, suggested as a TypeScript best practice

georgestagg:
* Switch from using `enum` for `RType` and others in favour of JS objects.
 * Switch to using lowercase property names.
 * Tweak the `RObjectPtr` structure to better match structure of `RObjectRaw` and `RObjectErr`.
* Change the `type` property name on `RTargetObj` to be named `targetType`, to avoid confusion with `RType`.
 * Make `RObjImpl.type()` a standard method rather than a computed property.
 * Tweak and simplify the R object duck typing tests to take into account the changes.

Many changes are just renaming such as `type` -> `targetType` and uppercase to lowercase. Apologies for not separating those out further into their own commit - it only just occurred to me.

### DIFF
--- a/src/webR/robj.ts
+++ b/src/webR/robj.ts
@@ -26,7 +26,7 @@ export type RFunction = RProxy<RObjFunction> & ((...args: unknown[]) => Promise<
 
 export type RPtr = number;
 
-export const RType = {
+export const RTypeMap = {
   null: 0,
   symbol: 1,
   pairlist: 2,
@@ -55,8 +55,8 @@ export const RType = {
   free: 31,
   function: 99,
 } as const;
-type RTypeKey = keyof typeof RType;
-type RTypeNumber = typeof RType[keyof typeof RType];
+export type RType = keyof typeof RTypeMap;
+export type RTypeNumber = typeof RTypeMap[keyof typeof RTypeMap];
 
 export const RTargetType = {
   raw: 0,
@@ -71,7 +71,7 @@ export type RTargetRaw = {
 
 export type RTargetPtr = {
   obj: {
-    type?: RTypeKey;
+    type?: RType;
     ptr: RPtr;
     methods?: string[];
   };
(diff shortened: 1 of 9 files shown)


src/webR/robj.ts
```diff
@@ -24,60 +24,67 @@ export type RRaw = RProxy<RObjRaw>;
...
-}
-
 export type RPtr = number;
 
-export enum RTargetType {
-  RAW = 'RAW',
-  PTR = 'PTR',
-  ERR = 'ERR',
-}
+export const RType = {
```
lionel-, line 29: `RTypeMap`? And then `RTypeKey` would be `RType`?

src/webR/robj.ts
```diff
@@ -24,60 +24,67 @@ export type RRaw = RProxy<RObjRaw>;
...
+  function: 99,
+} as const;
+type RTypeKey = keyof typeof RType;
+type RTypeNumber = typeof RType[keyof typeof RType];
+
+export const RTargetType = {
+  raw: 0,
+  ptr: 1,
+  err: 2,
+} as const;
```
lionel-, line 65: I can see how mapping the R types to their corresponding C enum value is useful, but here do we have any utility in using numbers? Should we use `type RTargetType = 'raw' | 'ptr' | 'err'` instead?

src/webR/robj.ts
```diff
@@ -24,60 +24,67 @@ export type RRaw = RProxy<RObjRaw>;
...
 
-export type RTargetPtr = {
-  obj: RPtr;
-  methods: string[];
-  type: RTargetType.PTR;
-};
 export type RTargetRaw = {
   obj: RawType;
-  type: RTargetType.RAW;
+  targetType: typeof RTargetType.raw;
```
lionel-, line 69: Then if I'm not mistaken this could be:

```suggestion
  targetType: 'raw';
```

lionel- (approved):

PR #87, opened 2022-11-18
Add `setNames()` method for setting the names attribute on an R Object

georgestagg:
Includes #86.

Note: Individually setting names elements with `Module._SET_STRING_ELT` is replaced in favour of using a `RObjCharacter` constructor in the next linked PR.

### DIFF
--- a/src/webR/robj.ts
+++ b/src/webR/robj.ts
@@ -237,12 +237,35 @@ export class RObjImpl {
     return RObjImpl.wrap(Module._ATTRIB(this.ptr)) as RObjPairlist;
   }
 
-  names(): (string | null)[] | null {
-    const attrs = this.attrs();
-    if (attrs.isNull()) {
-      return null;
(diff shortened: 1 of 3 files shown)


src/webR/robj.ts
```diff
@@ -240,12 +240,36 @@ export class RObjImpl {
...
+      values.forEach((v, i) => {
+        if (v === null) {
+          Module._SET_STRING_ELT(ptr, i, Module.getValue(Module._R_NaString, '*'));
+        } else {
+          const str = Module.allocateUTF8(String(v));
+          Module._SET_STRING_ELT(ptr, i, Module._Rf_mkChar(str));
+          Module._free(str);
+        }
+      });
+      Module._Rf_unprotect(1);
```
lionel-, line 258: Note: Could also protect `RObjImpl.null` in the first branch. This way there's no need to unprotect here.

lionel- (approved):

PR #88, opened 2022-11-18
Expand support for creating R objects from the main thread

georgestagg:
Includes #87 

This PR expands support for creating R objects from the main thread using `webR.newRObject()`. Creating lists, pairlists, atomic vectors, and environments is supported.

Creating new R objects from JS objects passed to `webR.newRObject()` in the form of an `RObjectTree<T>` (i.e. the type of the result obtained by running `RObjImpl.toTree()`) is supported, enabling round trip conversions.

Atomic vectors can also be created from JS scalars or arrays containing identical types. Lists and pairlists can similarly be created from JS arrays with mixed types.

`RProxy` objects can also be passed to `webR.newRObject()` to be included in lists or environments.

The `RObjImpl` subclass constructors each create an R object of that type from a compatible JS object or `RTargetPtr ` (derived from an `RProxy` passed from the main thread). Creating an object using the general `RObjImpl` constructor instead invokes `newRObjFromTarget()` to try and choose the best type of object to create, and then invokes that type's subclass constructor.


src/tests/webR/webr-main.test.ts
```diff
@@ -109,10 +118,134 @@ describe('Evaluate R code', () => {
...
+    expect(await rObj.toJs()).toEqual(
+      expect.objectContaining({
+        values: [{ re: 1, im: 2 }, { re: -3, im: -4 }, null],
+        names: null,
+        missing: [false, false, true],
+      })
+    );
+  });
+
+  test('Create a list', async () => {
```
lionel-, line 168: I have to say I find that sort of type inference a bit suspicious. I wish the syntax was a bit more unambiguous.

An alternative would be to treat the JS `[` syntax as a shortcut for `base::c()`, and let R deal with the task of combining disparate values. The outcome won't be better (e.g. deparsing to strings), but at least it'll be predictable for people who know about R's coercion rules.

In that case, I think this means:

- `[1, 2]` would create an integer vector rather than a double vector, as in base R.
- `null` would always represent the logical `NA`.
- We'd have a separate mechanism for creating lists, just like in R. Maybe we could have a simple ctor that adds a property to an array to mark it as a list? In this sort of arrays, `null` would represent `NULL`, and you'd use `[null]` to add a logical `NA`.

src/tests/webR/webr-main.test.ts
```diff
@@ -109,10 +118,134 @@ describe('Evaluate R code', () => {
...
+    });
+    expect(list.values[2]).toEqual({
+      type: 'character',
+      names: null,
+      values: ['abc'],
+      missing: [false],
+    });
+  });
+
+  test('Create an environment', async () => {
```
lionel-, line 193: I'm not sure it's going to be very useful to create environments from JS. On the other hand, being able to create data frames would be very useful since that's an important data type in R.

So I'm thinking we could use the object syntax for data frames, using recycling rules either of base R or the tidyverse (the latter is less error prone).

src/tests/webR/webr-main.test.ts
```diff
@@ -109,10 +118,134 @@ describe('Evaluate R code', () => {
...
+      type: 'environment',
+      names: ['newRObj', 'rObj'],
+      values: [newRObj, rObj],
+    })) as REnvironment;
+    const identical = (await webR.evalRCode('identical(rObj, newRObj)', env)).result as RLogical;
+    expect(await rObj.type()).toEqual('list');
+    expect(await identical.toLogical()).toEqual(true);
+  });
+
+  test('Round trip convert R object containing proxies and ensure result is identical', async () => {
```
lionel-, line 232: Nice!

src/tests/webR/webr-main.test.ts
```diff
@@ -109,10 +118,134 @@ describe('Evaluate R code', () => {
...
+      )
+    ).result as RList;
+    const jsObj = await rObj.toTree({ depth: 1 });
+    const newRObj = (await webR.newRObject(jsObj)) as RList;
+    const env = (await webR.newRObject({
+      type: 'environment',
+      names: ['newRObj', 'rObj'],
+      values: [newRObj, rObj],
+    })) as REnvironment;
+    const identical = (await webR.evalRCode('identical(rObj, newRObj)', env)).result as RLogical;
```
lionel-, line 245: Note that even if we lose the syntax for creating environments, we could still perform this sort of pattern by passing a list as `envir` argument of `base::eval()`.

lionel- (commented):
I only did a high level review to launch a discussion regarding the type inference rules.

src/tests/webR/webr-main.test.ts
georgestagg: I think creating environments is more useful in webR than in normal use of R as a way to organise the environment ready for `evalRCode` to execute in, as in the example commented upon below.

There's slightly more work involved in creating an environment object rather than just creating a list to be used as the environment, but managing an environment object for this feels more natural to me. OTOH I can see that the latter would also work OK - it is what Pyodide does by having a `globals` option for `runPython` of Python type `dict`.

Originally I had things set up so that a JS object becomes an environment, then IIRC I changed it to become list by default, but I like the idea of reserving the object syntax for data frames instead. I will remove the code that checks for JS objects in favour of reserving them for later use.

So for environments, you'll need to pass in `RObjectTree` type JS objects of the form`{ type: 'environment', names: [...], values: [...] }`, where you have to explicitly ask for the `environment` type. I think this is OK, as it going through the `isRObjectTree(obj)` test and `RObjEnvironment` constructor, rather than any specific conversion rule for creating environment objects.

georgestagg:
I've tweaked how `newRObjFromTarget` decides what type of R objects to create so that when there's ambiguity it focuses mostly on creating atomic vectors. Creating lists/environments directly from raw JS arrays and objects has been removed from the function.

In the current version, things now work the following way:

 * JS scalar `null` is converted to R `NULL`.
 * JS scalars are converted to R atomic vectors of length 1, based on the JS type.
 * JS arrays are converted to R atomic vectors by converting the arguments and then passing them through R's `c()`. This maintains the R coercion rules. In this case `null` is instead converted to R's logical `NA`.
 * JS objects of type `RObjectTree` are converted to the R object type as given in the `type` property, by passing the values to the relevant `RObject` constructor.

If a user really does want a list or environment object, it is still possible by passing in JS objects in the expanded `RObjectTree` form, e.g.
```
{ type: 'environment', names: ['foo'], values: [42] }
{ type: 'list', names: ['foo', 'bar'], values: [1, 'a'] }
```
Later, constructors can be added on the main thread to take in JS arrays and/or objects and output JS objects as above with the correct shape, if the above turns out to be too much friction for quick interactive creation of R lists.

Either way, I agree that we should have a future PR expand `RObjectTree` to support enough attributes to create a data frame (probably arbitrary attributes), and add such a data frame constructor on the main thread. I also like the idea of reserving the JS object as a quick way to create data frames.

src/tests/webR/webr-main.test.ts
georgestagg: > [1, 2] would create an integer vector rather than a double vector, as in base R.

A minor thing, but I think `c(1,2)` creates a double vector, and you have to explicitly ask for integers?

```
> class(c(1,2))
[1] "numeric"
> class(c(1L,2L))
[1] "integer"
```

So perhaps the closest behaviour is to convert `[1,2]` to double vector, unless someone specifically asks for integer with `{ type: 'integer', names: null, values: [1,2] }`.

lionel-:
Sounds good regarding list and environment creation.

However I have some concerns regarding the treatment of `null`. If we treat scalar JS objects as R vectors, then it seems it'd be more consistent to treat `null` as `NA`, since that's what it means in arrays.

>  In this case null is instead converted to R's logical NA.

`NULL` has length zero so it's not converted to `NA` by `c()` but ignored. This is another reason why it'd seem more consistent overall to treat `null` as logical `NA`.

src/tests/webR/webr-main.test.ts
lionel-: Yup that makes sense.

georgestagg:
Yes, fair enough argument with `null`. The current version explicitly changes the conversion behaviour when the `null` is given inside a JS array to replace it with logical `NA` rather than `NULL`, exactly because of the reason that it is ignored in `c()`

 I can certainly see the benefit of being consistent with other scalar types and converting `null` to logical `NA` overall, though. Probably I would separate parts of this function out into a new `newAtomicVectorFromTarget` function to make it fully clear what is going on.

However, with that change things will require a little further thought for someone who actually wants a `NULL`, as at the moment running `toTree()` on `RObjNull` just returns a raw JS `null`. Thinking about it, that is probably wrong and instead could instead be changed to something like
```
{ type: 'null', names: null, values: null }
```
or even just,
```
{ type: 'null' }
```
This could then be used here to create an actual R `NULL` if someone really wants it?

Alternatively, we simply set things up so that you can't create an R `NULL` with this method, and instead we add some other constructor that always just returns the result of `RObjImpl.null()`?

lionel-:
That's a tough one because we'll also want to make it easy to pass `NULL` arguments since it's a common sentinel for getting the default in R. I guess providing a ctor as you suggest would be good enough.

Since `NULL` is not a vector, I think `{ type: 'null' }` is the right representation.

georgestagg:
With the latest commit, when using `newRObjFromTarget`
* JS `null` always becomes logical `NA`.
* `RObjNull.toTree()` returns `{ type: 'null' }`.
* Explicitly passing in `{ type: 'null' }` to  creates an R `NULL`.

src/webR/robj.ts
```diff
@@ -478,6 +478,25 @@ export class RObjSymbol extends RObjImpl {
...
 
 export class RObjPairlist extends RObjImpl {
+  constructor(val: RawType | (RawType | null)[] | RTargetPtr | RObjectTree<RTargetObj>) {
+    if (isRTargetObj(val)) {
+      super(val);
+      return this;
+    }
+    const values = isRObjectTree(val) ? val.values : Array.isArray(val) ? val : [val];
+    const list = RObjImpl.wrap(Module._Rf_allocList(values.length)) as RObjPairlist;
+    list.preserve();
```
lionel-, line 488: Who is releasing the object? Should this be protect + unprotect at exit?

Ah I see below that preserving is intentional. What is the expected memory management scheme here? Is this where we'll hook in the shelter that we talked about?

src/webR/robj.ts
```diff
@@ -767,6 +825,22 @@ abstract class RObjAtomicVector<T extends atomicType> extends RObjImpl {
...
 export class RObjLogical extends RObjAtomicVector<boolean> {
+  constructor(val: RTargetObj | RObjAtomicData<boolean>) {
+    if (isRTargetObj(val)) {
+      super(val);
+      return this;
+    }
+    const values = isRObjectTree(val) ? val.values : Array.isArray(val) ? val : [val];
+    const ptr = Module._Rf_protect(Module._Rf_allocVector(RTypeMap.logical, values.length));
+    values.forEach((v, i) =>
+      Module.setValue(Module._LOGICAL(ptr) + 4 * i, v === null ? RObjImpl.naInt : Boolean(v), 'i32')
```
lionel-, line 836: Should probably use an lgl NA alias even if the underlying representation is int NA. This is what the C API of R does.

```suggestion
      Module.setValue(Module._LOGICAL(ptr) + 4 * i, v === null ? RObjImpl.naLgl : Boolean(v), 'i32')
```

src/tests/webR/webr-main.test.ts
```diff
@@ -109,10 +118,128 @@ describe('Evaluate R code', () => {
...
+    expect(await rObj.type()).toEqual('complex');
+    expect(await rObj.toJs()).toEqual(
+      expect.objectContaining({
+        values: [{ re: 1, im: 2 }, { re: -3, im: -4 }, null],
+        names: null,
+      })
+    );
+  });
+
+  test('Create a list', async () => {
```
lionel-, line 165: Looks like the tests still need to be updated?

src/webR/robj.ts
```diff
@@ -394,32 +377,44 @@ export class RObjImpl {
...
   static get null(): RObjNull {
-    return new RObjNull(Module.getValue(Module._R_NilValue, '*'));
+    return RObjImpl.wrap(Module.getValue(Module._R_NilValue, '*')) as RObjNull;
+  }
+
+  static get naInt(): number {
+    return Module.getValue(Module._R_NaInt, 'i32');
+  }
+
+  static get naReal(): number {
```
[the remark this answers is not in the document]
georgestagg: If we have `naInt` and `naLgl`, should we have `naDbl` to match? Could also expand to `naInteger`, `naLogical` and `naDouble`.

src/webR/robj.ts
georgestagg: Thanks! I'll pull out `LOGICAL()` and the other vector block data pointer lookups into a separate variable outside the loop.

src/webR/robj.ts
```diff
@@ -120,63 +121,53 @@ type RObjectTreeImpl<T> = {
...
+  // R's built in coercion rules
   if (Array.isArray(obj)) {
-    // Create a vector of logicals
-    const robjVec = Module._Rf_protect(Module._Rf_allocVector(RTypeMap.logical, obj.length));
-    obj.forEach((el, idx) => Module.setValue(Module._LOGICAL(robjVec) + 4 * idx, el, 'i32'));
-    Module._Rf_unprotect(1);
-    return RObjImpl.wrap(robjVec);
+    const objs = obj.map((el) => newRObjFromTarget({ targetType: 'raw', obj: el }));
+    const cString = Module.allocateUTF8('c');
+    const call = RObjImpl.protect(
```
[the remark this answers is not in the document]
georgestagg: If we're on the worker thread we can create an R pairlist from a JS array using the constructor directly,
```
> (new RObjPairlist([true, 3.14, 'hello'])).toArray()
< (3) [RObjLogical, RObjDouble, RObjCharacter]
```
For the main thread I guess we could also add something to avoid needing to define pairlists in the fully serialised form using `{ type: 'parilist', names: [...], values: [...] }`.

I think it's probably better to revisit that later when we're also looking at creating other objects like data frames, though. I expect in the end there'll be multiple functions to create objects from the main thread alongside this more general `newRObject` function, such as `newREnvironment`, `newRPairlist`, `newRDataframe` etc, each added independently in their own PR.

PR #89, opened 2022-11-18
Add `define()` method on `RObjEnvironment` that adds a value to an environment.

georgestagg:
Includes #88.


lionel- (approved):
Could also be `bind()` to follow rlang terminology.

PR #90, opened 2022-11-18
First version of webR Documentation

### LINKED ISSUE #41: More detailed documentation of the webR API is needed

georgestagg:
Documentation draft website at: https://docs.r-wasm.org/

Closes #41.


Anurodhyadav:
@georgestagg  The output.type for any r code gives { type: 'stdout', data: string } when i use  const output = await webR.read(); . I am not getting { type: 'canvasExec', data: string }, which i need to plot the graph when i run the following r code.

```
temperatures <- c(22, 27, 26, 24, 23, 26, 28)

result <- barplot(temperatures, 
  main = "Maximum Temperatures in a Week")

print(result).
```


The documentation for plotting the graph in webR `https://webr.quarto.pub/docs/communication.html#plotting-with-the-webr-html-canvas` is unclear for me. Could you please help me with this?

georgestagg:
There are probably multiple items in the output queue and you are getting only the first one. There will always be some introductory R output in the queue after initialisation. I will try to update the docs to make that clearer.

You could try using [`webR.flush()`](https://webr.quarto.pub/docs/api/js/classes/WebR.WebR.html#flush) to recieve all the waiting output messages at once in an `Array`, or run `await webR.read()` in an asyncronous loop and deal with each item as they arrive. Alternatively, you could use the built in [`Console`](https://webr.quarto.pub/docs/evaluating.html#console ) class to manage a simple asynchronous loop for you. When using the `Console` class the HTML canvas element can be accessed with the [Console.canvas](https://webr.quarto.pub/docs/api/js/classes/WebR.Console.html#canvas) instance property.

If you are still having issues after trying the above steps, please feel free to open a new GitHub issue, just so that the comments here remain about docs in general.

Anurodhyadav:
Thanks alot @georgestagg . I was able to make it work using Console class

PR #92, opened 2022-11-22
Don't add `missing` property during conversion of atomic vectors with `toTree()`

georgestagg:
At the moment the default is to use `toArray()` to populate the `values` property when converting atomic vectors to JavaScript. Since `toArray()` returns `null` for missing values unambiguously, there is no need to include the `missing` property as part of the result.

The `missing` property remains optional on the `RObjectTree` type in preparation for a possible future option introduced so that `toTree()` returns a JavaScript typed array for the values property instead, which would include the raw sentinel values and so indeed require a separate array of missing indicators.

### DIFF
--- a/src/webR/robj.ts
+++ b/src/webR/robj.ts
@@ -762,7 +762,6 @@ abstract class RObjAtomicVector<T extends atomicType> extends RObjImpl {
       type: this.type(),
       names: this.names(),
       values: this.toArray(),
-      missing: this.detectMissing(),
     };
   }
 }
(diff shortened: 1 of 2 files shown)


src/webR/robj.ts
```diff
@@ -762,7 +762,6 @@ abstract class RObjAtomicVector<T extends atomicType> extends RObjImpl {
       type: this.type(),
       names: this.names(),
       values: this.toArray(),
-      missing: this.detectMissing(),
```
lionel-, line 765: I suspect some tests will need adjustments?

georgestagg: It looks like `.missing` was not properly covered to begin with, with the atomic vector tests just checking `.values` directly. Probably those tests were modified versions of older tests from before switching to the tree structure with `.names` and `.missing` added.

I'll update those tests to include checking `names == null` and `type` is correct.

georgestagg: Ah! I'll probably catch that one in a rebase then :)

PR #96, opened 2022-11-23
Turn off wasm decoding

lionel-:
The webR build supports setting `WEBR_REPO` to a local repo. When set, the repo is preloaded with `--preload-file` so that webR starts with packages already installed.

This works fine except for packages with C++ libraries that dynamically initialise global variables using the R API, e.g. a top-level:

```cpp
SEXP sym = Rf_install("foo")
```

dplyr does this for instance. When this happens and `noWasmDecoding` isn't set, Emscripten instantiates these variables before R has started, causing a crash.

### DIFF
--- a/src/webR/webr-worker.ts
+++ b/src/webR/webr-worker.ts
@@ -353,6 +353,12 @@ function init(config: Required<WebROptions>) {
   Module.noAudioDecoding = true;
   Module.noInitialRun = true;
 
+  // Don't instantiate .so libraries packaged through `WEBR_REPO` too
+  // early. Otherwise C++ libraries with dynamic initialisation of
+  // global variables might call into the R API too early, beforeR has
+  // started.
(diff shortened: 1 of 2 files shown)


src/webR/webr-worker.ts
```diff
@@ -353,6 +353,12 @@ function init(config: Required<WebROptions>) {
   Module.noAudioDecoding = true;
   Module.noInitialRun = true;
 
+  // Don't instantiate .so libraries packaged through `WEBR_REPO` too
+  // early. Otherwise C++ libraries with dynamic initialisation of
+  // global variables might call into the R API too early, beforeR has
```
georgestagg, line 358: Minor typo
```suggestion
  // global variables might call into the R API too early, before R has
```

georgestagg (approved):
Good catch!

PR #97, opened 2022-11-24
Add hook for `system()` calls

lionel-:
Useful to skip tests using `system()` in testthat:

```r
options(
  webR.hook_system = function(cmd, ...) {
    if (nzchar(Sys.getenv("TESTTHAT"))) {
      testthat::skip("Can't call `system()` in webR.")
    }
    c("", "")
  }
)
```


georgestagg (approved):

lionel-:
@georgestagg I've added a scaffholding for running R tests implemented in the webr package. Could you please review the last two commits, which also add unit tests for the functionality implemented here?

src/tests/webR/webr-extra-pkgs.test.ts
```diff
@@ -0,0 +1,21 @@
...
+});
+
+beforeAll(async () => {
+  await webR.init();
+});
+
+test('webr package tests succeed', async () => {
+  // TODO: Use the simpler eval functions that converts R errors to JS
+  // exceptions, this way we'll get information about failing tests
+  const out = (await webR.evalR('webr:::self_test()')).result;
```
georgestagg, line 15: I guess in the new scheme this would simply be something like,

```
const out = await webR.evalR('webr:::self_test()');
expect(await out.type()).toEqual('null');
```

lionel-:
Regarding the tests locations, just to be clear, the current tests, while using the webr package, are not testing the webr package. They are testing webR itself. So I opted for webR/ instead of packages/ when considering the location. Does this make sense?

lionel-:
I guess the file should be called `webr-r.test.ts` though if we keep the current location.

georgestagg:
Yes, that makes sense. I has assumed `webr-extra-pkgs.test.ts` would be expanded later with other self tests, in particular for code inside `webr` as in R package. However, I can see that this file is testing webR itself, so does make sense in its current location.

I support the renaming to `webr-r.test.ts`, although I appreciate that it's not obvious naming this file since it doesn't really correspond to any matching TypeScript `.ts` file.

PR #98, opened 2022-11-24
Fix timezone support

lionel-:
Without these changes:

```r
# `$zone` component is not set to "ACST"
as.POSIXlt("2013-06-17 22:33:44", tz="Australia/Darwin")

# Returns `NA`
Sys.timezone()
```

WIP: I'd like to add unit tests, but I haven't been able to run them locally yet.

### DIFF
--- a/R/Makefile
+++ b/R/Makefile
@@ -146,7 +146,8 @@ $(BUILD)/state/r-stage2-configured: $(BUILD)/state/r-patched $(FORTRAN_WASM_LIB)
 	    --disable-nls \
 	    --enable-byte-compiled-packages=no \
 	    --enable-static=yes \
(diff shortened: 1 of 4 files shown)


georgestagg (approved):
Looks good to me!

Re: running unit tests locally, if you have rebuilt `R.bin.js` you might need to re-run the sed in https://github.com/georgestagg/webR/blob/fb0afdbcb2ee329221978ab6ec7789d80d9955e5/Makefile#L17 before the unit tests will run. When unit testing webR is loaded with a relative directory as the `WEBR_URL` and so triggers the bug.

I've been meaning to move that out into a different makefile target somewhere so that it will be done automatically when running `make check` from inside the `src` dir.

PR #99, opened 2022-11-28
Add `libpng` to `wasm/lib`

lionel-:
So that packages from webr-repo can link to it.

Using `EM_LIBS = -s USE_LIBPNG=1` only works in `CFLAGS` for include files. It is ignored in `LDFLAGS` because Emscripten doesn't link side modules to system libs or ports.

I tried directly linking to ports in the Emscripten cache but this causes compilation issues because the ports are not compiled with `-fPIC`. So unfortunately it looks like we can't use Emscripten ports for CRAN packages, unless we link them in the main lib with `EMCC_FORCE_STDLIBS=` to work around dead-code elimination. However that would increase the binary size even in cases where the libs are not needed.

So instead, this PR adds a `libs` directory to webR to maintain our own ports for system libs. It isn't part of the default build. Instead it should be built manually before running `make` in webr-repo.

Alternatively, we could build several versions of webR with different sets of system libs statically linked in. This would have the advantage of reducing the total size when several packages link to the same lib. But we can think about this when/if this problem comes up.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 /dist
 /host
+/libs/build
+/libs/download
 /wasm
 /tools/fortran.mk
 /tools/dragonegg/build
--- a/Makefile
+++ b/Makefile
@@ -9,19 +9,17 @@ EMFC_FILES = $(EMFC) $(FORTRAN_WASM_LIB)
 
 # Build webR and install the web app in `./dist`
 .PHONY: webr
-webr: $(EMFC_FILES)
+webr: $(EMFC_FILES) libs
 	cd R && $(MAKE) && $(MAKE) install
 	cd src && $(MAKE)
 
 $(EMFC_FILES):
 	cd $(EMFC_DIR) && $(MAKE) && $(MAKE) install
 
-
-# Supporting libs for packages of webr-repo
+# Supporting libs for webr
 .PHONY: libs
 libs:
 	cd libs && $(MAKE)
-
 
 # Build webR in a temporary docker container
 .PHONY: docker-webr
--- a/Makefile
+++ b/Makefile
@@ -44,13 +44,18 @@ check-pr:
 	cd src && $(MAKE) lint && $(MAKE) check && $(MAKE) check-packages
 
 .PHONY: clean
-clean: clean-webr
+clean: clean-wasm
 	cd tools/dragonegg && $(MAKE) clean
 	cd tools/flang && $(MAKE) clean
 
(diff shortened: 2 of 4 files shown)


georgestagg:
Just a quick comment to say that if we do go own this route we should probably move the libpcre and libxz ports that are currently built in `R/Makefile` to this `Makefile` instead, and then link to the resulting static libraries as part of the webR build process.

georgestagg:
Strangely enough I've found that adding `-lpng` to the package linking flags works for me without having to rebuild libpng for Emscripten independently, in that it seems to statically link a version of `libpng.a` compiled for wasm and with `-fPIC` enabled built by Emscripten.

I added `-Wl,--trace` to the `LDFLAGS` to work out where the static library was coming from:
```
[...]/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/pic/libpng.a
```

I still think something like this PR is worth doing, though, for two reasons.
 * Not all system libraries will be available as Emscripten ports
 * We might be able to get dynamic linking working with downloading of wasm compiled `.so` side modules on demand.

The latter should work for libraries under 4kb in size and avoid the increased package download sizes that comes with static linking.

georgestagg:
> the Makevars in png uses `libpng-config --static --ldflags` which expands to `-lpng16` which Emscripten doesn't provide?

Ah, yes, I had removed using `libpng-config` from png's `Makevars` file for experimentation. I had assumed we'd need to edit the `Makevars` in any case, since running `libpng-config` returns the flags that are required to link to system (well, homebrew in my case) `libpng`, rather than the `libpng` compiled for WASM.

It doesn't look like there's an `libpng16.a` in that cache directory, but I think using `libpng.a` works.

> I'd still like to introduce a way of supplying custom Makevars in webr-repo without having to patch the source package though.

Yes, this is a good idea. I've already had to patch several package provided `Makevars` for building with Emscripten. A more formal way to do this makes a lot of sense.

georgestagg (approved):
Sorry it took me so long to come back to this proposal.

I've been experimenting with building the cairo library for webr, so that the built in bitmap graphics devices work better (ultimately, for shiny support). As you predicted in the OP, the Emscripten ports version of `libpng` causes linking issues, and this PR fixes them.

So, we should rebase and merge. My main suggestion is that we also move the XZ and PCRE build rules here, and so make building the new `libs` directory part of the default build. I have already done that as part of an experimental cairo branch, so I am happy to merge as is and make that change in a future PR.

PR #103, opened 2022-12-05
Backport R's strict-prototypes fix for dotcall functions

georgestagg:
R source has recently committed a fix for the strict-prototypes problem with R's dotcall functions. Previously this broke running under Emscripten and was worked around for webR with a patch to use a JavaScript trampoline, avoiding the function pointer casting to incompatible signatures.

This PR contains the fix in R source, backported to the R-4.1.3 patches for webR. The workaround adding a trampoline has been removed. With the extra call into JS and a function lookup into `wasmTable` removed, performance and stack use should be improved for R workloads with many calls into compiled functions.

~Even better, once the change hits a released version of R we can upgrade R and avoid this patch entirely.~

Thanks to @wch for prompting me to look into this again.


georgestagg:
Oh drat! R does indeed start up and seem to work with this new patch. However, the issue is still partially present - at least for some R functions. Disappointing, but a win for solid test schemes!

R's `rnorm` (a wrapper around `.Call(C_rnorm, ...)`) now works fine with this patch, with the `.Call` succeeding without issue. However, the following R code crashes:

```
x = runif(10)
y = x^3 + runif(10, min = -0.1, max = 0.1)
df <- data.frame(x, y)
m <- nls(y ~ I(x^power), data = df, start = list(power = 1))
```

Closing this PR for the moment. I will investigate and if there is a way forward with this I'll re-open.

georgestagg:
...and I think we're back in action!

The backported fix from R is insufficient. It does not properly take into account Fortran symbol prototypes and so whilst `do_dotcall` is now fixed, the function `do_dotCode` continues to crash when running under Emscripten. This explains why `rnorm` worked OK, but other R functions did not.

In particular, part of R's strict-prototypes fix casts Fortran functions to be of the same type as C functions, which look like

```
SEXP fun(SEXP, [...], SEXP);
```

However, Fortran subroutines should actually look more like,
```
void fun(void**, [...], void**);
```
since they do not return anything and take in arguments as references.

So, I have added a patch on top of the backported R code to include a further 65 `typedef`s for use with Fortran symbols, with the name `FFUNSX` to match R's `FUNSX`, where `X` is the number of arguments. In an ideal world this would all be done with a set of pre-processor macros, but this way matches R's implementation.

When it comes to upgrading the version of R built with webR we should confirm if this continues to be an issue and report it to the R devel team if so. IIUC they have successfully silenced the strict-prototypes compiler error, but including this extra patch on top would also make the fix more portable.

PR #105, opened 2022-12-08
Protect R objects in `evalJs()`

lionel-:


src/webR/webr-worker.ts
```diff
@@ -396,10 +396,14 @@ function init(config: Required<WebROptions>) {
...
         );
-        const call = Module._Rf_lang2(Module._Rf_install(stop), Module._Rf_mkString(msg));
+
+        const ffi_msg = Module._Rf_protect(Module._Rf_mkString(msg));
+        const call = Module._Rf_protect(Module._Rf_lang2(Module._Rf_install(stop), ffi_msg));
         Module._free(stop);
         Module._free(msg);
+
         Module._Rf_eval(call, RObjImpl.baseEnv.ptr);
+        Module._Rf_unprotect(2);
```
georgestagg, line 406: Can this line be reached? I thought `stop` ultimately ended in a `longjmp` to top level. If I am incorrect, perhaps the whole dance with `stop` can be removed and replaced with a direct call to `Module._Rf_error()`?

lionel-: oh got it regarding stop/Rf_error. You're right this line shouldn't be reachable. To be extra clear, we can replace it by a `stopUnreachable()` JS function meant to catch situations that shouldn't happen.

This makes me think that we might need to be careful not to throw from top-level, only inside a function or `.Call()` context, so that the protection count can be reset on error?

lionel-: This line is now replaced by a call to a new helper `throwUnreachable()`.

lionel-:
You'd still need to protect the result of `mkChar()` here:

```js
Module._SET_STRING_ELT(errMsg, 0, Module._Rf_mkChar(msg));
```

So the current setup seems better.

georgestagg (approved):

PR #106, opened 2022-12-08
Don't create JS stubs for main module exports

lionel-:
Simple change but it brings R.bin.js from 15-20mb to 900-1500kb by removing all the JS wrappers around the main module exports, in particular the functions from the Fortran lib.


georgestagg (approved):
Very nice!

`DECLARE_ASM_MODULE_EXPORTS` doesn't seem to be documented at https://emscripten.org but I can see it's well established and documented in `settings.js`.

It sounds like WASM R must be running in the global scope for this to work. I think that should be OK, even if we want multiple instances running in the future we would probably want a separate web worker per R instance anyway, and each can run in their own worker's global scope.

PR #107, opened 2022-12-08
Rename evalR to captureR, create a simpler version for evalR, re-throw R conditions as exceptions

georgestagg:
A new option to `captureR` is also added so that captured R error conditions can be re-thrown as JS exceptions. This setting is on by default.

The unit tests are also updated to reflect the new naming and the way R errors are thrown.

### DIFF
--- a/src/tests/webR/webr-worker.test.ts
+++ b/src/tests/webR/webr-worker.test.ts
@@ -1,5 +1,5 @@
 import { WebR } from '../../webR/webr-main';
-import { RInteger, RLogical, RRaw, RList, RCharacter } from '../../webR/robj';
+import { RInteger, RLogical, RRaw } from '../../webR/robj';
 
 const webR = new WebR({
   WEBR_URL: '../dist/',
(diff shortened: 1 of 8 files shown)


src/webR/webr-worker.ts
```diff
@@ -320,14 +349,27 @@ function evalR(code: string, env?: RTargetPtr, options: EvalROptions = {}): RTar
...
+    const error = (output.toArray() as RObjImpl[]).find(
+      (out) => out.get('type').toString() === 'error'
+    );
+    if (error) {
+      throw new Error(
+        error.pluck('data', 'message')?.toString() || 'An error occured evaluating R code.'
+      );
+    }
+  }
+  Module._Rf_unprotect(2);
```
lionel-, line 367: I think this is a top-level case where we might leak objects. For instance, we never unprotect before throwing an error.

This is a sensitive part of the code so here is a strategy that might be more manageable:

- Wrap the whole logic in try/finally and call unprotect in the finally branch
- Increment a protection counter at each protect call.

src/tests/webR/webr-worker.test.ts
```diff
@@ -52,19 +52,13 @@ describe('Test webR virtual filesystem', () => {
...
 
   test('Raise a JS exception as an R condition', async () => {
-    const res = (await webR.evalR('webr::eval_js("51+")')).output as {
-      type: string;
-      data: RList;
-    }[];
-    expect(res[0].type).toEqual('error');
-    const cndMessage = (await res[0].data.get('message')) as RCharacter;
-    expect(await cndMessage.toString()).toContain('Unexpected end of input');
+    const badSyntax = webR.evalR('webr::eval_js("51+")');
```
lionel-, line 60: 🎉

lionel- (approved):

src/tests/webR/webr-worker.test.ts
georgestagg: I love how this is going through several layers of R/JS evaluating and the exception survives now!

src/webR/webr-worker.ts
georgestagg: Oh of course! I should have noticed that!

If I had spotted it I would have saved `error.pluck('data', 'message')?.toString()` into a JS object and then add an unprotect(2) before the throw. I like `finally` better though to avoid duplication.

georgestagg: I've wrapped the entire function in a try-finally block, and keep track of the number of protects using `protectCounter`, finally unprotecting in the finally branch.

I also added a comment explaining the need to keep track of  protect use, since I can easily see myself forgetting or simply not noticing later.

PR #108, opened 2022-12-09
Send output to console.log and console.warn during basic evalR use

georgestagg:
Since `evalR` now does more than just unwrapping `result`, a new dispatch type is added so that the logging to console of stream output and conditions only needs to be implemented once on the worker side. The `result` value is passed back to the main thread, reproducing the previous behaviour.

### DIFF
--- a/src/tests/webR/webr-main.test.ts
+++ b/src/tests/webR/webr-main.test.ts
@@ -89,7 +89,35 @@ describe('Evaluate R code', () => {
     expect((await webR.read()).data).toBe('Hello, stderr!');
   });
 
-  test('Capture stdout while evaluating R code', async () => {
+  /* Since console.log and console.warn are called from the worker thread in the
+   * next two tests, we cannot mock them in the usual way. Instead we spy on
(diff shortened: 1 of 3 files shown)


src/tests/webR/webr-main.test.ts
```diff
@@ -89,23 +89,42 @@ describe('Evaluate R code', () => {
...
+    expect(buffer.includes('[1]  30  42  66  70  78 102')).toEqual(true);
+  });
+
+  test('Send conditions to console.warn while evaluating R code', async () => {
+    const spyStderr = jest.spyOn(process.stderr, 'write');
+    await webR.evalR('warning("This is a warning!")');
+    const buffer = spyStderr.mock.calls[0][0] as Buffer;
+    expect(buffer.includes('Warning message: \nThis is a warning!')).toEqual(true);
+  });
+
```
lionel-, line 110: Can we also test that R errors are converted to JS ones?

lionel- (approved):

PR #111, opened 2022-12-13
Construct R objects on the main thread by proxying the worker thread's RObjImpl class constructor

georgestagg:
This PR tweaks the syntax for how new R objects are created on the main thread. Rather than using `webR.newRObject([...])`, the R object class is proxied on `webR` and the constructor function is hooked as part of the proxy. Construction is intercepted and converted into a `newRObject` message to be sent to the worker thread.

The result is that new R objects are more explicitly created using the syntax,
```
const obj = await new webR.RObject([...])
```

The `await` is a little strange but required due to the communication channel.

Creating a new R object is otherwise similar to before this PR, where serialised `RObjectTree` types are converted to the relevant R object type, and JS arrays are converted into R atomic vectors according to the rules of R's `c()` function.

This change is partial progress towards #112 in which this scheme is expanded so that different types of R object types can be constructed on the main thread, by exposing the different R object class constructors through proxies.

### DIFF
--- a/src/webR/webr-main.ts
+++ b/src/webR/webr-main.ts
@@ -1,7 +1,7 @@
 import { newChannelMain, ChannelMain, ChannelType } from './chan/channel';
 import { Message } from './chan/message';
 import { BASE_URL, PKG_BASE_URL } from './config';
-import { newRProxy } from './proxy';
+import { newRProxy, DistProxy } from './proxy';
 import { unpackScalarVectors, replaceInObject } from './utils';
 import {
   RTargetObj,
@@ -9,10 +9,10 @@ import {
   isRObject,
   RawType,
   RList,
-  RObjectTree,
-  NamedObject,
   REnvironment,
   RCharacter,
+  RObjData,
+  RObjImpl,
 } from './robj';
 
 export type CaptureROptions = {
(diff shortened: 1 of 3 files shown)


src/webR/webr-main.ts
```diff
@@ -188,13 +192,18 @@ export class WebR {
     }
   }
 
-  async newRObject(
-    jsObj: RawType | RawType[] | RObjectTree<RObject> | NamedObject<RawType | RObject>
-  ): Promise<RObject> {
-    const targetObj = replaceInObject(jsObj, isRObject, (obj: RObject) => obj._target);
+  #newRObjConstructor<T, R>() {
```
georgestagg, line 195: Note: The type parameters here are used later in #112 for proxying other R object constructor functions to the main thread that take different argument types and return different subclasses of `RObject`.

src/webR/webr-main.ts
```diff
@@ -60,12 +60,16 @@ const defaultOptions = {
   channelType: ChannelType.Automatic,
 };
 
+type RData = DistProxy<RObjData>;
```
lionel-, line 63:
```suggestion
type RData = DistProxy<RObjData>;

```

lionel- (approved):
Nice!

PR #112, opened 2022-12-13
Construct R objects of specific type on the main thread by proxying further R object class constructors

georgestagg:
Includes #111.

The aim of this PR is to make available several R object constructors on the main thread, for creating specific types of R object from potentially ambiguous JS objects.

The following R object constructors are proxied and exposed on the main thread, in the same way as the `RObjImpl` constructor is exposed as `webR.RObject` in (#111):
 - `RNull`
 - `RLogical`, `RInteger`, `RDouble`, `RComplex`, `RCharacter`, `RRaw`
 - `RList`, `RPairlist`
 - `REnvironment`

In addition, several R object constructor functions on the worker thread have been tweaked so that they can additionally accept objects of the type e.g. `NamedObject<X>`. This type corresponds to JavaScript objects with keys of type `string` and values of type `X`.

This change allows for specific types of R object to be created from the main thread, without having to know about automatic coercion or type mapping rules built into webR or R itself.

Invoking a specific R type constructor in this way feels much more robust than my previous attempts to handle conversion of ambiguous items (such as JS objects) into R objects. For example, with this change either an R list or vector can be created from a JS object, with named values, from the main thread, by invoking the relevant proxied R object constructor.

```
const list = await new webR.RList({ a: 1, b: 2, c: 3 });
const arr = await new webR.RInteger({ a: 1, b: 2, c: 3 });
```

Since we know in advance that `webR.RList` returns an object of type `list`, the calling user also no longer needs to add the type information for TypeScript with `as RList`, unlike when using the current method with `webR.newRObject()` which always returns a generic `RObject`.

The previous `webR.newRObject()` behaviour still exists via the `webR.RObject` constructor. For this, the result does not have a specific associated R object type and its type is encoded as the dummy R type named 'object' in the message to the worker thread. The worker thread detects 'object' and invokes the generic `RObjImpl` constructor in that case.

It is my intention that `webR.RObject` remains as a constructor that automatically "does the right thing", and it is the constructor that should be used when creating objects from serialised form (e.g. from the result of `toJs()`). As with before this change, it also accepts JS scalars and arrays to create an atomic vector of type according to the rules of `c()`. JS objects are not accepted in `webR.RObject`, even though they now work in other constructors, because as previously discussed we might use the JS object syntax for automatically converting into data frames in the future.

WIP: The typing around R tree objects needs to be tweaked to be more consistent and support serialising other R types. This will be done in a separate PR.

### DIFF
--- a/src/tests/webR/webr-main.test.ts
+++ b/src/tests/webR/webr-main.test.ts
@@ -243,7 +243,7 @@ describe('Create R objects using serialised form', () => {
   });
 });
 
-describe('Create R atomic vectors from JS arrays', () => {
+describe('Create R vectors from JS arrays using RObject constructor', () => {
   test('Convert a JS null to R logical NA', async () => {
     const rObj = (await new webR.RObject(null)) as RLogical;
     expect(await rObj.type()).toEqual('logical');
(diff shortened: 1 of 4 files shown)


src/tests/webR/webr-main.test.ts
```diff
@@ -299,18 +299,183 @@ describe('Create R atomic vectors from JS arrays', () => {
...
+      expect.objectContaining({
+        values: Object.values(jsObj),
+        names: Object.keys(jsObj),
+      })
+    );
+  });
+
+  test('Create a list containing both a logical NA and R NULL', async () => {
+    const rNull = await new webR.RNull();
+    const jsObj = [true, 2, null, rNull];
```
georgestagg, line 383: Note: JS `null` continues to encode an R logical NA. A real R `NULL` can be obtained by adding the result of,
```
await new webR.RNull()
```
to the input array.

Alternatively, a real R NULL can be obtained from serialised form with,

```
const rNull = await new webR.RObject({ type: 'null' });
```

src/tests/webR/webr-main.test.ts
```diff
@@ -299,18 +299,183 @@ describe('Create R atomic vectors from JS arrays', () => {
   });
 });
 
+describe('Create R objects from JS objects using proxy constructors', () => {
+  test('Create an R NULL', async () => {
+    const rObj = await new webR.RNull();
```
lionel-, line 304: This seems strange to me because `NULL` is a singleton. Can we cache `RNull` in `webR` instead? Or maybe in a property where we'd collect cached values, constants, and singletons, for example a cached `TRUE` object? These objects would be populated at init time and wouldn't be waited on by users.

Maybe `webR.objs.RNull` in this case.

src/webR/robj.ts
```diff
@@ -1194,3 +1205,25 @@ export function getRObjClass(type: RTypeNumber): typeof RObjImpl {
...
+
+/*
+ * Normalise the various types of acceptable RObjData into consistently typed
+ * arrays of names and values.
+ */
+function normaliseRObjData<T>(jsObj: RObjAtomicData<T>): {
+  names: (string | null)[] | null;
+  values: (T | null)[];
+};
+function normaliseRObjData(jsObj: RObjData) {
```
lionel-, line 1217: I tend to avoid verbs like "normalise" or "standardise" or "process" in favour of more descriptive ones.

In this case maybe this can simply be `as` (or `to`, not sure what is more common in JS)? So `asRObjData()`

lionel- (approved):
Looks great!

src/tests/webR/webr-main.test.ts
georgestagg: Yes, I like this idea a lot. I like how it covers other objects like `TRUE`. I think the objects would still have to be proxies, but the IIUC the proxy targets shouldn't change so we should be OK to cache such objects.

I will remove this implementation of RNull from the PR and add a new one later for cached objects.

src/webR/robj.ts
georgestagg: I like `toRObjData()`, thanks :)

PR #114, opened 2022-12-21
Add R object cache to main thread for long-living R objects

georgestagg:
A continuation of #112.

The R object class proxying function, now named `newRObjClassProxy`, has been expanded to enable transfer of static properties of `RObjImpl`. This is done by adding a `getStaticPropertyValue()` static method to `RObjImpl` and tweaking the `callRObjMethod` request type to optionally invoke static methods. The function is also moved to the `proxy.ts` file as part of the change, so as to match where the similar R object instance proxying function, `newRProxy`, lives.

(Note: I considered proxying the static properties directly, but this way better matches the R object instance proxies where only methods are hooked, to avoid potentially confusing situations where it's not clear that requesting a property with `obj.foo` does something complicated and must be `await`ed.)

A similar `getPropertyValue()` method is also added to `RObjImpl`, replacing the need for a separate `getRObjProperty` request type containing mostly repeated code and only used in one place, which has now been removed.

The new feature is used to create an R object cache on the main thread at `webR.objs`. Objects that are constant for the duration of the session (e.g. R's `NULL`) can be added to this cache at init time by grabbing them from static methods on the `RObjImpl` class, where properties such as `RObjImpl.null` already existed.

I have added a few objects to get things started: `NULL`, `TRUE`, `FALSE`, logical `NA`, the base env, and the global env. I'm happy to add any others if there are obvious but missing R objects here.

I really like that this allows for neat tricks such as easily accessing base env symbols,
```
> const ver = await webR.objs.baseEnv.pluck('R.version', 'version.string');
> await ver.toString()
< 'R version 4.1.3 (2022-03-10)'
```
or binding variables to the global environment from JS in one line,
```
await webR.objs.globalEnv.bind('abc', 123);
```

This also provides another way to clearly differentiate between `NULL` and logical `NA` when constructing lists,
```
const list = await new webR.RList([webR.objs.na, webR.objs.null])
```

### DIFF
--- a/src/tests/webR/webr-main.test.ts
+++ b/src/tests/webR/webr-main.test.ts
@@ -10,6 +10,7 @@ import {
   RPairlist,
   REnvironment,
   RInteger,
+  RFunction,
 } from '../../webR/robj';
 
 const webR = new WebR({
@@ -378,8 +379,7 @@ describe('Create R objects from JS objects using proxy constructors', () => {
   });
 
   test('Create a list containing both a logical NA and R NULL', async () => {
-    const rNull = await new webR.RObject({ type: 'null' });
-    const jsObj = [true, 2, null, rNull];
+    const jsObj = [true, 2, null, webR.objs.null];
     const rObj = await new webR.RList(jsObj);
     expect(await rObj.type()).toEqual('list');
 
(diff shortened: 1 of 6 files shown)


src/tests/webR/webr-main.test.ts
```diff
@@ -493,6 +492,45 @@ describe('Serialise nested R lists, pairlists and vectors unambiguously', () =>
...
+    expect(await webR.objs.false.toLogical()).toEqual(false);
+  });
+
+  test('Logical NA', async () => {
+    expect(await webR.objs.na.type()).toEqual('logical');
+    expect(await webR.objs.na.toLogical()).toEqual(null);
+  });
+
+  test('R global environment', async () => {
+    await webR.objs.globalEnv.bind('abc', 123);
```
lionel-, line 517: Clean up after this test?

lionel- (approved):
Nice!

PR #115, opened 2023-01-05
Fix build on macOS Ventura and latest Emscripten

lionel-:
- Don't build with Cocoa to work around issues with newer SDKs. It also makes the build faster and it's not necessary.
- Set the stack size for compatibility with newer Emscripten versions where the size is set to 64kb.


R/Makefile
```diff
@@ -120,6 +121,7 @@ MAIN_LDFLAGS  = -s MAIN_MODULE=1
 MAIN_LDFLAGS += -s WASM=1
 MAIN_LDFLAGS += -s WASM_BIGINT
 MAIN_LDFLAGS += -s ALLOW_MEMORY_GROWTH=1
+MAIN_LDFLAGS += -s STACK_SIZE=1MB
```
georgestagg, line 124: This currently fails in GHA because it's using Emscripten 3.1.14 (where the option is named `TOTAL_STACK` instead) and I've been putting off updating the GHA docker image due to some issues building flang in an x86 VM on my machine.

I'm happy to commit, but I'll need to bring forward updating the GHA image to latest Emscripten before release.

PR #116, opened 2023-01-10
Rename type definition `RTargetObj` to `WebRPayload`

georgestagg:
Including general renaming of items with "target" to use "payload" throughout.

With this change we have,

`export type WebRPayload = WebRPayloadRaw | WebRPayloadPtr | WebRPayloadError`

### DIFF
--- a/src/webR/proxy.ts
+++ b/src/webR/proxy.ts
@@ -49,7 +49,7 @@ type RProxify<T> = T extends Array<any>
  *   - Where an RObjImpl would be returned, an RProxy is returned instead
  *   - All return types are wrapped in a Promise
  *
- * If required, the WebRPayload object associated with the proxy can be
+ * If required, the WebRPayloadPtr object associated with the proxy can be
  * accessed directly through the _payload property.
  */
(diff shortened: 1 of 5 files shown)


src/webR/proxy.ts
```diff
@@ -49,11 +49,11 @@ type RProxify<T> = T extends Array<any>
  *   - Where an RObjImpl would be returned, an RProxy is returned instead
  *   - All return types are wrapped in a Promise
  *
- * If required, the proxy target can be accessed directly through the _target
- * property.
+ * If required, the WebRPayload object associated with the proxy can be
```
lionel-, line 52:
```suggestion
 * If required, the WebRPayloadPtr object associated with the proxy can be
```

Since it doesn't make sense to wrap the other kinds of payload?

lionel- (approved):
Much better!

It's just a bit strange to have abbreviated `Ptr` and unabbreviated `Error` suffixes. Should it be `WebRPayloadPointer`? Or `WebRPayloadErr`?

PR #117, opened 2023-01-11
Safety improvements to `captureR()` and `evalR()`

lionel-:
- Protect the result of `captureR()` in `evalR()` before calling `get()` which allocates.
- Deallocate C strings in `finally` clause of `captureR()` to avoid leaks.
- New helper type `DictStrings` to help with the above.
- Some style improvements (paragraphs, shorter name for `nProt`, use of prefix `++` instead of postfix to more easily see the side effect).


georgestagg (approved):
Looks great, thanks!

Let's rebase and merge this after the big `RObject` and `WebRData` reorg, since we'll have to deal with some renamed types.

PR #119, opened 2023-01-11
Use idiomatic cleanup-unwind and env introspection

lionel-:
- Add `on_exit()` wrapper of `on.exit()` that sets `add = TRUE, after = FALSE` for safe registration of dtors. Use `on_exit()` to clean up more safely.

- Take `env = parent.frame()` and use that. We generally prefer taking the direct caller env as soon as possible, save it in a variable, and then use that, rather than taking indirect caller envs with a number of frames to skip. The latter is more brittle to changes in the code.

### DIFF
--- a/packages/webr/R/eval.R
+++ b/packages/webr/R/eval.R
@@ -17,8 +17,12 @@
 #'
 #' @export
 #' @useDynLib webr, .registration = TRUE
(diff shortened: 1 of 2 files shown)


georgestagg (approved):
Great, thanks! I like the `on_exit` utility function.

PR #121, opened 2023-01-12
Tweaks to how length 1 atomic vectors are converted to JS scalars

georgestagg:
* Rename `toLogical` to `toBoolean`.
* Convert length 1 vectors to pure JS scalar types - e.g. `toBoolean()` returns `boolean` rather than `boolean | null`.

In the case of a length 1 atomic vector with a missing value of `NA`, an error is thrown.

For JS numbers, there is an alternative to throwing an error on `NA`: return `NaN` instead. This may be more natural as the JS `NaN` value counts as a pure `number` type, and since it is equal to the `NA` sentinel for doubles it has a simple implementation (simply return the first value from `toTypedArray()` rather than `toArray()`).

The downside is that everywhere else we use `null` rather than `NaN` to signal a missing value. In theory I think webR should treat them the same, but nevertheless I've kept things consistent for the moment and throw an error.

### DIFF
--- a/src/webR/robj-worker.ts
+++ b/src/webR/robj-worker.ts
@@ -780,15 +780,15 @@ export class RLogical extends RVectorAtomic<boolean> {
     super({ payloadType: 'ptr', obj: { ptr } });
   }
 
-  getLogical(idx: number): boolean | null {
+  getBoolean(idx: number): boolean | null {
     return this.get(idx).toArray()[0];
   }
 
-  toLogical(): boolean | null {
+  toBoolean(): boolean | null {
     if (this.length !== 1) {
       throw new Error('Unable to convert atomic vector of length > 1 to a scalar JS value');
     }
-    return this.getLogical(1);
+    return this.getBoolean(1);
   }
 
   toTypedArray(): Int32Array {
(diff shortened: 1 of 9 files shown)


src/webR/robj-worker.ts
```diff
@@ -780,15 +780,19 @@ export class RLogical extends RVectorAtomic<boolean> {
...
 
-  toLogical(): boolean | null {
+  toBoolean(): boolean {
     if (this.length !== 1) {
       throw new Error('Unable to convert atomic vector of length > 1 to a scalar JS value');
     }
-    return this.getLogical(1);
+    const val = this.getBoolean(1);
+    if (val === null) {
+      throw new Error('Unable to convert missing value `NA` to a JS boolean');
```
lionel-, line 793: Should we use the tidyverse error style guide? It stipulates either "can't" or "must" phrasing. So here it would be "Can't convert ..."

lionel- (approved):

src/webR/robj-worker.ts
georgestagg: Thanks. A quick `grep` returned a few other instances of "Unable to" so I've replaced those too.

PR #123, opened 2023-01-13
Move Emscripten utils to common `emscripten.ts` file

lionel-:
Branched from #122.


src/webR/emscripten.ts
```diff
@@ -109,3 +109,13 @@ export interface Module extends EmscriptenModule {
 }
 
 export const Module = {} as Module;
+
+type EmPtr = ReturnType<typeof Module.allocateUTF8>;
```
georgestagg, line 113: Nice use of `ReturnType`, I think right now this is just a `number` but I suppose in theory it could change if Emscripten decides to implement something different.

I wonder if we should use the same trick to define the `RPtr` type?

georgestagg (approved):

src/webR/emscripten.ts
lionel-: hmm yes, maybe taking `EmPtr` instead of `RPtr` would make it clearer what's going on? I.e. that we're taking the equivalent of a C pointer.

georgestagg: Yes. It would be a change in line with the `RObject` refactor were we switched away from the `RObj` prefix to a `WebR` prefix for types not strictly to do with R itself.

PR #124, opened 2023-01-13
Add debugging tools to inspect addresses of R objects

lionel-:
For both the R and JS sides.

Also add `parseEvalBare()` helper to safely and easily evaluate R code with inputs. The `Bare` suffix anticipates the addition of a higher level `parseEval()` function that calls `evalR()`.


georgestagg (approved):
Looks good to me, will come in useful when debugging strange memory issues!

PR #125, opened 2023-01-13
Add shelter stack to preserve R objects on the JS side

lionel-:
- When an R object is transferred from the worker to the main thread, it gets automatically added to a shelter. A shelter is a dynamic list that protects all its elements.

- Shelters are arranged in a stack. WebR users call `webR.shelter.push()` to create a new shelter before doing some work with R objects. When they are done, they call `webR.shelter.pop()` to unprotect at once all objects transferred to the main thread during this time. This should usually be done in a `finally` clause to avoid leaks on jumps.

- If the user forgot to push a shelter before transferring objects to the main thread, webR automatically creates a new shelter and a warning is issued in the JS console.

In the future I'd like to explore a more automatic way of sheltering objects sent to the main thread, as part of a simplification of channel payloads. For instance this would simplify how conditions in the output array are transmitted by `captureR()`.


src/repl/repl.ts
```diff
@@ -26,7 +26,7 @@ function resizeTerm() {
   (async () => {
     await webR.init();
     const dims = fitAddon.proposeDimensions();
-    webR.evalR(`options(width=${dims ? dims.cols : 80})`);
+    webR.evalR(`options(width=${dims ? dims.cols : 80})`, undefined, false);
```
lionel-, line 29: The evaluation methods get a new `shelter?: boolean` argument to disable sheltering. This is mostly for internal usage. For instance here it avoids having to push and pop a shelter when we know we don't need one.

src/tests/webR/webr-main.test.ts
```diff
@@ -109,7 +145,7 @@ describe('Evaluate R code', () => {
     spyStdout.mockRestore();
   });
 
-  test('Send conditions to console.warn while evaluating R code', async () => {
+  test.skip('Send conditions to console.warn while evaluating R code', async () => {
```
lionel-, line 148: As discussed on Slack, this test fails for some reason.

src/webR/robj-worker.ts
```diff
@@ -410,18 +420,25 @@ export class RPairlist extends RObject {
...
-    list.preserve();
-    for (
-      let [i, next] = [0, list as Nullable<RPairlist>];
-      !next.isNull();
-      [i, next] = [i + 1, next.cdr()]
-    ) {
-      next.setcar(new RObject(values[i]));
+    list.protect();
+
+    try {
```
lionel-, line 428: While I was adjusting these methods to remove the `preserve()` calls, I took the opportunity to move cleanups to `finally` clauses for safety.

src/webR/webr-main.ts
```diff
@@ -160,48 +165,49 @@ export class WebR {
...
-        const output: any[] = [];
-        for await (const out of outList) {
-          const type = await ((await out.pluck(1, 1)) as RCharacter).toString();
-          const data = await out.get(2);
-          if (type === 'stdout' || type === 'stderr') {
-            output.push({ type, data: await (data as RString).toString() });
-          } else {
-            output.push({ type, data });
+
+      case 'raw': {
```
lionel-, line 184: Since I refrained from adding a sheltering method on the main side, I had to move the handing of output items to the worker side so I could shelter them from there. This has the side benefit of removing a good deal of channel communication just to unwrap output messages.

src/webR/webr-main.ts
```diff
@@ -218,4 +224,42 @@ export class WebR {
       }
     }
   }
+
+  async evalNumber(code: string, env?: REnvironment, shelter?: boolean): Promise<number> {
```
lionel-, line 228: This helper (and the upcoming variants `evalString()` etc) could be improved in the future by sending raw payloads to avoid unnecessary channel communications.

src/webR/webr-worker.ts
```diff
@@ -349,7 +429,7 @@ function captureR(code: string, env?: WebRPayloadPtr, options: CaptureROptions =
   }
 }
 
-function evalR(code: string, env?: WebRPayloadPtr): RObject {
+function evalR(code: string, env?: RPtr): RObject {
```
lionel-, line 432: I changed the type of `env` from a payload to a bare pointer to make it easier to call from the worker side.

src/repl/repl.ts
georgestagg: If `shelter` is mainly for internal usage, is it worth moving it to live inside `CaptureROptions` to match the deep options `withHandlers`, `throwJsException` etc.

The downside is that `evalR` will always use the default value for `shelter`, but if it is important you can always switch to the more advanced `captureR` to gain access to the `options` object?

PR #126, opened 2023-01-16
Always return `WebRPayload` when making a channel request

georgestagg:
With this commit, `ChannelMain.request` is modified to always return a `WebRPayload`. This formalises what we were pretty much doing anyway.

With this change, we also move the `try`-`catch` blocks that capture errors thrown on the worker thread higher up the stack, wrapping the entire branch of `dispatch()` that deals with requests.

The change provides three immediate benefits,
 * Avoids multiple type casting to `WebRPayload` throughout,
 * Avoids repeated code,
 * Any errors that occur during a webR request are captured and handled, rather than in just some branches and not others.

This fixes an issue where some webR errors during a request (in particular, FS errors) break the worker's main R loop. By always capturing the errors and sending them to the main thread to be dealt with, the  main R loop continues and webR can recover gracefully.

With this PR FS errors are not properly handled on the main thread, but the FS code is being refactored in a future linked PR in any case.

### DIFF
--- a/src/webR/chan/channel.ts
+++ b/src/webR/chan/channel.ts
@@ -4,6 +4,7 @@ import { ServiceWorkerChannelMain, ServiceWorkerChannelWorker } from './channel-
 import { WebROptions } from '../webr-main';
 import { isCrossOrigin } from '../utils';
 import { IN_NODE } from '../compat';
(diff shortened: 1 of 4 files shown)


lionel- (approved):
Makes sense!

Issue #56, opened 2022-08-26
Persistent filesystem storage in the webR REPL


georgestagg:
Perhaps by mounting an Emscripten `IDBFS` filesystem somewhere like `/storage`.

We will need to think about long term storage implications. Browser localStorage will work in the short term but could be cleared away at any moment. Maybe integrating with third party storage like Dropbox via user auth is an option.

georgestagg:
Copying a comment here with an update on the current situation,

For webR in particular a persistent VFS is tricker than at first glance. We cannot immediately use solutions that work well for other WASM projects because we use a blocking worker thread to maintain our place in the stack while waiting for input for R. Due to this we cannot easily yield to the event loop in the worker thread[1].

From what I currently understand there are several options available, but each have their own drawbacks for us,

localStorage and sessionStorage are unavailable in web workers, where we perform our R computation.
"OPFS" looks promising, but at the time of writing is only available in bleeding-edge Chromium browsers.
Emscripten's new "wasmFS" also looks promising but is currently Work in Progress.
The IndexedDB API (used by Emscripten's IDBFS driver) is asynchronous only. Since we block in the worker thread we're not able to use this API.
Service workers can be (ab)used to persist data, but only a single service worker can be active on the page at a time.
I think the future of persistent storage in webR is going to be based around synchronous OPFS or wasmFS API, once they are ready and supported by more browsers. In the meantime, solving this issue will probably involve significant new infrastructure based around passing messages to the main thread to persist data.

[1] Asyncify would allow us to yield, but it does not work well for us due to the way the R interpreter works and the large asyncify overhead induced.

HTUser-1:
@georgestagg , since this post, OPFS is wide supported in more browser. Wordpress playground already implemented it: https://github.com/WordPress/wordpress-playground/pull/548 
Also, the new JSPI : https://v8.dev/blog/jspi can be an alternative for asyncify overhead issues.
Best,

aehlke:
It looks unclear whether Safari on Mac or iOS will get it anytime soon :/ They have it in no preview state and I see no plans for it. Unfortunately I think OPFS isn't a universal solution.

seanbirchall:
Might be useful I made this with claude code for the SAB channel

https://github.com/seanbirchall/webr-opfs

PR #127, opened 2023-01-16
Expand and reorganise webR virtual filesystem API

georgestagg:
Requires #126.

Expands and reorganises the webR filesystem API,
 * Move functions into a `FS` namespace.
 * Add new `mkdir`, `rmdir`, and `unlink` functions.
 * Rename existing functions to match Emscripten FS API.
 * Re-throw any error payloads returned.

### DIFF
--- a/src/webR/webr-main.ts
+++ b/src/webR/webr-main.ts
@@ -131,17 +131,6 @@ export class WebR {
       await this.#chan.request(msg);
     }
   }
-  async putFileData(name: string, data: Uint8Array) {
-    const msg = { type: 'putFileData', data: { name: name, data: data } };
-    return await this.#chan.request(msg);
-  }
-  async getFileData(name: string): Promise<Uint8Array> {
-    return (await this.#chan.request({ type: 'getFileData', data: { name: name } }))
-      .obj as Uint8Array;
-  }
-  async getFSNode(path: string): Promise<FSNode> {
-    return (await this.#chan.request({ type: 'getFSNode', data: { path: path } })).obj as FSNode;
-  }
 
   async captureR(
     code: string,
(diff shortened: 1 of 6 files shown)


src/webR/webr-main.ts
```diff
@@ -218,4 +207,64 @@ export class WebR {
...
+      if (payload.payloadType === 'err') {
+        const e = new Error(payload.obj.message);
+        e.name = payload.obj.name;
+        e.stack = payload.obj.stack;
+        throw e;
+      }
+    },
+    unlink: async (path: string): Promise<void> => {
+      const payload = await this.#chan.request({ type: 'unlink', data: { path } });
+      if (payload.payloadType === 'err') {
```
georgestagg, line 262: There is now quite a lot of repeated code to check for `WebRPayloadErr` and re-throw the error. When we reorganise the `WebRPayload`s as a class hierarchy with constructors we should be able to mitigate this by including a method to throw the `WebRPayloadErr` class.

georgestagg: For the moment I've extracted the repeated `Error` construction into a new `webRPayloadError()` function.

lionel- (commented):
I only took a quick look but LGTM.

htiuser:
When working on this part, please don't forget : https://github.com/georgestagg/webR/issues/56
Since this request, Persistent Emscripten VFS is more mature. For example, SQLite is supporting several ways: https://sqlite.org/wasm/doc/trunk/persistence.md
It will also be a must to support this new SQLite Wasm on WebR.
Best,

georgestagg:
For webR in particular a persistent VFS is tricker than at first glance. We cannot immediately use solutions that work well for other WASM projects because we use a blocking worker thread to maintain our place in the stack while waiting for input for R. Due to this we cannot easily yield to the event loop in the worker thread[1].

From what I currently understand there are several options available, but each have their own drawbacks for us,

* `localStorage` and `sessionStorage` are unavailable in web workers, where we perform our R computation.
* "OPFS" looks promising, but at the time of writing is only available in bleeding-edge Chromium browsers.
* Emscripten's new "wasmFS" also looks promising but is currently Work in Progress.
* The IndexedDB API (used by Emscripten's `IDBFS` driver) is asynchronous only. Since we block in the worker thread we're not able to use this API.
* Service workers can be (ab)used to persist data, but only a single service worker can be active on the page at a time.

I think the future of persistent storage in webR is going to be based around synchronous OPFS or wasmFS API, once they are ready and supported by more browsers. In the meantime, solving #56 would probably involve significant new infrastructure based around passing messages to the main thread to persist data. That amount of work should wait until after the first tagged version of webR has been released.

[1] Asyncify would allow us to yield, but it does not work well for us due to the way the R interpreter works and the large asyncify overhead induced.

htiuser:
> For webR in particular a persistent VFS is tricker than at first glance. We cannot immediately use solutions that work well for other WASM projects because we use a blocking worker thread to maintain our place in the stack while waiting for input for R. Due to this we cannot easily yield to the event loop in the worker thread[1].
> 
> From what I currently understand there are several options available, but each have their own drawbacks for us,
> 
> * `localStorage` and `sessionStorage` are unavailable in web workers, where we perform our R computation.
> * "OPFS" looks promising, but at the time of writing is only available in bleeding-edge Chromium browsers.
> * Emscripten's new "wasmFS" also looks promising but is currently Work in Progress.
> * The IndexedDB API (used by Emscripten's `IDBFS` driver) is asynchronous only. Since we block in the worker thread we're not able to use this API.
> * Service workers can be (ab)used to persist data, but only a single service worker can be active on the page at a time.
> 
> I think the future of persistent storage in webR is going to be based around synchronous OPFS or wasmFS API, once they are ready and supported by more browsers. In the meantime, solving #56 would probably involve significant new infrastructure based around passing messages to the main thread to persist data. That amount of work should wait until after the first tagged version of webR has been released.
> 
> [1] Asyncify would allow us to yield, but it does not work well for us due to the way the R interpreter works and the large asyncify overhead induced.

Well understood! Important for us is to have this persistent file storage in the future!

PR #130, opened 2023-01-18
Add subclass wrappers

lionel-:
I spent some time trying to create some sort of generic static method that would automatically pick up the type of subclasses of `RObject`, but I couldn't make it work.

### DIFF
--- a/src/webR/robj-worker.ts
+++ b/src/webR/robj-worker.ts
@@ -46,7 +46,7 @@ function newObjectFromData(obj: WebRData): RObject {
     const objs = obj.map((el) => newObjectFromData(el));
     const cString = Module.allocateUTF8('c');
     const call = RObject.protect(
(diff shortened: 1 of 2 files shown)


georgestagg:
Does the following work?

```
  static wrap<T extends typeof RObject>(this: T, ptr: RPtr): InstanceType<T> {
    const type = Module._TYPEOF(ptr);
    return new (getRWorkerClass(type as RTypeNumber))({
      payloadType: 'ptr',
      obj: { ptr },
    }) as InstanceType<T>;
  }
```

Using [TypeScript `this` parameters](https://www.typescriptlang.org/docs/handbook/2/classes.html#this-parameters)?

lionel-:
oh no sorry I had something different (a global function not a static method taking `this`).

Let me try your idea.

georgestagg (approved):

PR #131, opened 2023-01-18
Update env methods

lionel-:
This PR updates the `new()` and `bind()` methods of `REnvironment` to fix memory leaks. It takes this opportunity to introduce and use wrappers of the `Module` entry points.

- Add the foundation for memory management on the worker side: `keep()` for implicit protection of objects sent to the main thread, `RObject.free()` for release on the main side, `protect()` and `unprotect()` wrappers for temporary protection on the worker side.

- `protect()` is generic and takes `RHandle = RObject | RPtr`. A new `handlePtr()` function helps extracting a pointer.

- Add `envPoke()` wrapper that takes `RHandle` inputs.

- Update the `new()` and `bind()` methods of `REnvironment` to use the above and fix potential leaks in the process.

- Add `string` constructor for symbols. Used to create symbols without having to manage the C string memory.

I plan to go over the rest of the file in a new PR.

### DIFF
--- a/src/webR/robj-worker.ts
+++ b/src/webR/robj-worker.ts
@@ -6,6 +6,16 @@ import { parseEvalBare } from './utils-r';
 import { isWebRDataTree, WebRDataTree, WebRDataTreeAtomic, WebRDataTreeNode } from './tree';
 import { WebRDataTreeNull, WebRDataTreeString, WebRDataTreeSymbol } from './tree';
 
+export type RHandle = RObject | RPtr;
+
+export function handlePtr(x: RHandle): RPtr {
+  if (isRObject(x)) {
+    return x.ptr;
+  } else {
+    return x;
+  }
+}
+
 export interface ToTreeOptions {
   depth: number;
 }
--- a/src/webR/robj-worker.ts
+++ b/src/webR/robj-worker.ts
@@ -375,6 +375,24 @@ export class RNull extends RObject {
 }
 
 export class RSymbol extends RObject {
+  // Note that symbols don't need to be protected. This also means
+  // that allocating symbols in loops with random data is probably a
+  // bad idea because this leaks memory.
+  constructor(x: string) {
+    if (typeof x !== 'string') {
+      super(x);
+      return;
+    }
+
+    const name = Module.allocateUTF8(x);
+
+    try {
+      super(RObject.wrap(Module._Rf_install(name)));
+    } finally {
+      Module._free(name);
+    }
+  }
+
   toTree(): WebRDataTreeSymbol {
     const obj = this.toObject();
     return {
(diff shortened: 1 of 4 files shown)


src/webR/robj-worker.ts
```diff
@@ -613,35 +664,51 @@ export class REnvironment extends RObject {
...
+      ++nProt;
+
+      values.forEach((v, i) => {
+        const name = names ? names[i] : null;
+        if (!name) {
+          throw new Error("Can't create object in new environment with empty symbol name");
+        }
+
+        const sym = new RSymbol(name);
+        const vObj = protect(new RObject(v));
```
lionel-, line 683: Protecting these objects before assignement is necessary because `defineVar()` allocates.

src/webR/robj-worker.ts
```diff
@@ -2,10 +2,28 @@ import { Module } from './emscripten';
...
+    return x.ptr;
+  } else {
+    return x;
+  }
+}
+
+// Use this for implicit protection of objects sent to the main
+// thread. Currently uses the precious list but could use a different
+// mechanism in the future. Unprotection is explicit through
+// `RObject.free()`.
```
georgestagg, line 22: Thanks for adding comments for this and `free()`, the separation would look strange without the explanation.

src/webR/robj-worker.ts
```diff
@@ -64,10 +82,14 @@ function newObjectFromData(obj: WebRData): RObject {
   throw new Error('Robj construction for this JS object is not yet supported');
 }
 
+// FIXME: Can we simplify this?
+// Do we need to take payloads?
```
georgestagg, line 86: I think this can be done, and with the recent refactoring I don't think we need to take payloads here anymore. The important part is being able to differentiate between creating a new `RObject` from some `WebRData`, and creating an `RObject` reference from an `RPtr` (needed for `RObject.wrap()`).

I agree that this can be handled simpler and I think a second constructor argument with a reasonable default could work to signal the latter situation, I'll investigate this.

src/webR/robj-worker.ts
```diff
@@ -112,18 +134,25 @@ export class RObject {
...
+  // TODO: Remove these
   protect(): void {
     this.ptr = Module._Rf_protect(this.ptr);
   }
-
   unprotect(): void {
     Module._Rf_unprotect_ptr(this.ptr);
   }
 
+  // TODO: Remove these
```
georgestagg, line 152: Since we're cleaning up `RObject` don't forget to remove the `static` versions of these methods too (`unprotect`, `unprotectPtr`, `preserveObject` and `releaseObject`), near the end of the class definition. This is the duplication I mentioned earlier. We'd ideally replace all these with the unscoped helper functions other than the new instance method `free()`.

georgestagg (approved):
LGTM and good work spotting leaks.

I'll spend a little time looking into simplifying `RObject` construction while keeping `RObject.wrap()` working.

src/webR/robj-worker.ts
lionel-: yup I plan to clean this up in a more encompassing PR. This one just lays out the groundwork.

PR #132, opened 2023-01-18
Fix protection issues in `newObjectFromData()`

lionel-:
Branced from #131 

* In the array case we recurse into `newObjectFromData()` but forgot to protect the resulting objects.

* I extracted the array method into its own function to make the main dispatch more symmetric.

* New `protectInc()` helper that makes it easier to manage protection counters.


src/webR/robj-worker.ts
```diff
@@ -57,29 +57,38 @@ function newObjectFromData(obj: WebRData): RObject {
...
+  throw new Error('Robj construction for this JS object is not yet supported');
+}
+
+// JS arrays are interpreted using R's c() function, so as to match
+// R's built in coercion rules
+function newObjectFromArray(arr: WebRData[]) {
+  const prot = { n: 0 };
+
+  try {
+    const objs = arr.map((el) => protectInc(newObjectFromData(el), prot));
```
lionel-, line 73: `protectInc()` is particularly nice here because we don't need to save the result to a variable just to call `++nProt`, which would force us to have a multiline lambda.

georgestagg (approved):

PR #133, opened 2023-01-19
Fix memory issues in remaining RObject methods

lionel-:
I moved the `keep()` calls into `RObject`'s base constructor. This is temporary, in the future `keep()` will only be called from the channel dispatcher.

### DIFF
--- a/src/webR/robj-worker.ts
+++ b/src/webR/robj-worker.ts
@@ -117,7 +117,9 @@ export class RObject {
     } finally {
       // FIXME: Shouldn't preserve in the constructor, only in channel
       // messages
-      keep(this.ptr);
+      if (this.ptr) {
+        keep(this.ptr);
+      }
     }
   }
 
--- a/src/webR/robj-worker.ts
+++ b/src/webR/robj-worker.ts
@@ -272,32 +272,23 @@ export class RObject {
   }
 
   set(prop: string | number, value: RObject | WebRDataRaw) {
-    let idx: RPtr;
-    let char: RPtr = 0;
-    if (typeof prop === 'number') {
-      idx = Module._Rf_protect(Module._Rf_ScalarInteger(prop));
-    } else {
-      char = Module.allocateUTF8(prop);
-      idx = Module._Rf_protect(Module._Rf_mkString(char));
+    const prot = { n: 0 };
+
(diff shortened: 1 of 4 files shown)


src/webR/robj-worker.ts
```diff
@@ -628,33 +639,47 @@ export class RList extends RObject {
...
   }
 }
 
 export class RString extends RObject {
+  // Unlike symbols, strings are not cached and must thus be protected
+  constructor(x: string) {
+    if (typeof x !== 'string') {
+      super(x);
+      return;
+    }
```
georgestagg, line 672: The TypeScript signature claims `x` is always `string` here. Should it be something like the following?
```suggestion
  constructor(x: RObject | string) {
    if (isRObject(x)) {
      super(x);
      return;
    }
```

src/webR/utils-r.ts
```diff
@@ -13,6 +13,28 @@ export function protectInc<T extends RHandle>(x: T, prot: { n: number }): T {
...
 
+export function protectWithIndex(x: RHandle): { loc: number; ptr: RPtr } {
+  // Integer size hardcoded to 4 bytes. This is fine but is there a
+  // way to call sizeof?
+  const pLoc = Module._malloc(4);
+  const loc = Module.getValue(pLoc, 'i32');
+
+  Module._R_ProtectWithIndex(handlePtr(x), pLoc);
+
+  return { loc: loc, ptr: pLoc };
```
georgestagg, line 24: I'm not quite sure what's happening here. If the location pointed to by `pLoc` is populated during `_R_ProtectWithIndex`, should the `Module.getValue()` line come after?

georgestagg (approved):
This looks really nice, thank you!

I think the use of your memory management helper functions and `new RObject(foo)` where foo is an `RObject` really helps tidy up the code a lot.

I've added a few minor comments but nothing serious.

src/webR/robj-worker.ts
lionel-: Good catch.

src/webR/utils-r.ts
lionel-: aha yes it should! I was wondering why `loc` would cause a crash!

Issue #128, opened 2023-01-16
Memory management in webR


lionel-:
# Current state

This section shows (approximatively, some tweaks are still needed) the current state of memory management in webR.

## Protecting and unprotecting singular objects

Objects must be protected from the worker thread, not from the main thread. Consider:

```typescript
async function foo() {
  const obj = await webR.evalR('rnorm(1)');
  await obj.keep();
}
```

Because of concurrency, R code may run between the two `await` expressions, causing `obj` to be GC'd before it has a chance to be protected. We don't have any choice but to automatically preserve all objects returned to the main thread. The `keep()` method must be called implicitly by the worker.

On the other hand, because Javascript doesn't have reliable object finalisers, `free()` can't be called implicitly. If the main thread doesn't explictly free these objects, memory is leaked:

```typescript
async function foo() {
  const obj = await webR.evalR('rnorm(1)');

  return await bar(obj);
}

async function bar(obj: RObject) {
  const other = await webR.evalR('runif(1)');
  const out = await webR.evalR('x + y', { x: obj, y: other});

  return await out.toNumber();
}
```

Code on the main thread must protect-unwind object release in `finally` clause:

```typescript
async function foo() {
  // Implicit `keep()`
  const obj = await webR.evalR('rnorm(1)');

  // Need to protect-unwind `free()` as soon as created
  try {
    return await bar(obj);
  } finally {
    obj.free();
  }
}
```

`try / finally` can quickly get out of hands with this singular unprotection approach though:

```typescript
async function bar(obj: RObject) {
  const other = await webR.evalR('runif(1)');

  try {
    const out = await webR.evalR('x + y', { x: obj, y: other});

    try {
      return await out.toNumber();
    } finally {
      out.free();
    }
  } finally {
    other.free();
  }
}
```

To avoid this increasing nestedness we need some way of unprotecting multiple objects at once.

## Evaluating for side effects

Implicit protection gets in the way when evaluating R code for side effects. min the following example the return value is leaked because the user forgot to explictly release it.

```typescript
async function printNorm(): void {
  await webR.evalR('print(rnorm(1))');
}
```

The user needs to remember to release the returned object immediately with a rather heavy boilerplate:

```typescript
async function printNorm(): void {
  await (await webR.evalR('print(rnorm(1))')).free();
}
```

Alternatively we could add arguments to `evalR()`, `captureR()`, `newRObject()`, etc to turn off implicit protection:

```typescript
async function printNorm(): void {
[...]

PR #135, opened 2023-01-19
Consistently preserve objects before transmission

lionel-:
Branched from #134.
Part of #128.

- Move implicit protection from the `RObject` ctor to the channel dispatcher.

- The `captureR()` unwrapping of stdout/stderr outputs now happens on the worker side. This spares channel transmissions.

- Add `destroy()` method to `RObject` and remove other memory methods. I've called it `destroy()` because we are planning to disable objects that have been released, and this name makes it clear that the object can no longer be used after this method has been called.

### DIFF
--- a/src/tests/webR/robj.test.ts
+++ b/src/tests/webR/robj.test.ts
@@ -518,28 +518,21 @@ describe('Reject incorrect object types during R object construction', () => {
 });
 
 describe('Garbage collection', () => {
-  test.skip('Protect and release R objects', async () => {
+  test('Protect and release R objects', async () => {
     const gc = (await webR.evalR('gc')) as RFunction;
     await gc.exec(false, false, true);
     const before = await ((await gc.exec(false, false, true)) as RDouble).toTypedArray();
 
     const mem = await webR.evalR('rnorm(10000,1,1)');
-    mem.destroy();
-
     const during = await ((await gc.exec(false, false, true)) as RDouble).toTypedArray();
 
+    mem.destroy();
     const after = await ((await gc.exec(false, false, true)) as RDouble).toTypedArray();
 
     expect(during[0]).toBeGreaterThan(before[0]);
     expect(during[1]).toBeGreaterThan(before[1]);
 
-    // TODO: For some reason `exec()` causes this test to fail after
-    // switching to protect(). Also we now temporarily call `keep()`
-    // from `new RObject()` so there's bound to be leaks until this is
-    // moved to the channel (for instance this causes new symbols to
-    // be preserved).
     expect(after[0]).toBeLessThan(during[0]);
-
     expect(after[1]).toBeLessThan(during[1]);
   });
 });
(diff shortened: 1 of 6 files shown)


src/tests/webR/robj.test.ts
```diff
@@ -467,10 +467,10 @@ describe('Garbage collection', () => {
     const before = await ((await gc.exec(false, false, true)) as RDouble).toTypedArray();
 
     const mem = await webR.evalR('rnorm(10000,1,1)');
-    mem.preserve();
+    mem.destroy();
```
lionel-, line 470: Not quite right, I'll fix that.

georgestagg: OK, if the test can't be fixed yet feel free to leave this test as is and skipped until the work for #128 is complete.

georgestagg: As an experiment, I removed `keep()` in the constructor and then the following (while clunky) seemed to start working again,

```
    const gc = (await webR.evalR('gc')) as RFunction;
    await (await gc.exec(false, false, true)).destroy();

    let gcObj = (await gc.exec(false, false, true)) as RDouble;
    const before = await gcObj.toTypedArray();
    await gcObj.destroy();

    const mem = await webR.evalR('rnorm(10000,1,1)');

    gcObj = (await gc.exec(false, false, true)) as RDouble;
    const during = await gcObj.toTypedArray();
    await gcObj.destroy();

    await mem.destroy();

    gcObj = (await gc.exec(false, false, true)) as RDouble;
    const after = await gcObj.toTypedArray();
    await gcObj.destroy();

    expect(during[0]).toBeGreaterThan(before[0]);
    expect(during[1]).toBeGreaterThan(before[1]);
    expect(after[0]).toBeLessThan(during[0]);
    expect(after[1]).toBeLessThan(during[1]);
```

src/webR/webr-worker.ts
```diff
@@ -108,29 +110,87 @@ function dispatch(msg: Message): void {
...
+              } as WebRPayloadPtr;
+
+              write({
+                payloadType: 'raw',
+                obj: {
+                  result: resultPayload,
+                  output: output,
+                },
+              });
+            } catch (_e) {
```
georgestagg, line 173: The entire `'request'` branch of the `switch` is now surrounded by a `try/catch` that returns thrown errors as a webR error payload, so I think you can just use `try/finally` here and let any errors bubble up and be caught by that instead.

georgestagg (approved):
Thanks, this looks great! I like how you've moved the processing of `captureR` output to the worker thread to save transfers, it does make sense to have the processing there.

It looks like `keep(this.ptr)` is still part of the `RObject` constructor with this PR? I suppose that is why the memory management test continues to fail. Do we still need `keep()` there if we now manage memory on the worker with `protect()`, and run`keep()` during channel transfer? If not it would be good to remove it.

PR #136, opened 2023-01-20
Simply R object construction using new `RObjectBase` class for R object references

georgestagg:
Simplify the construction of `RObject` and its subclasses so as to not need to work with `WebRPayload` objects. This feels better since we're not working with transferring data over the channel here.

We need to discriminate between `WebRData` and `RPtr` during `RObject` construction, since we can't always use `wrap()` inside subclass constructors as it leads to an infloop.

A new `RObjectBase` class is added that just sets `obj.ptr` and does nothing else. Instances of `RObjectBase` are constructed and used to signal to the `RObject` constructors that we have an existing `RPtr`. During object construction what was `{ payloadType: 'ptr', obj: { ptr } }` now becomes `new RObjectBase(ptr)`, and `{ payloadType: 'raw', obj: foo }` just becomes `foo`.

`RObjectBase` is set as a super class to `RObject`. This helps when testing for "`RObject` or `RObjectBase`", in that with this hierarchy in place both can be tested for at once with `obj instanceof RObjectBase`. It also fits well in that an R object reference fundamentally is just a Wasm pointer.

The `isRObject()` implementation has also been simplified.

### DIFF
--- a/src/webR/robj-worker.ts
+++ b/src/webR/robj-worker.ts
@@ -1,5 +1,4 @@
 import { Module } from './emscripten';
-import { WebRPayload, isWebRPayload, isWebRPayloadPtr, isWebRPayloadRaw } from './payload';
 import { Complex, isComplex, NamedEntries, NamedObject, WebRDataRaw } from './robj';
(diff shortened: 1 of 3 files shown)


lionel- (approved):
👍

PR #137, opened 2023-01-20
Check object type in `RObject` subclass constructors and reject incorrect objects

georgestagg:
Based on #136.

Checks the object type during `RObject` subclass construction and rejects objects of incorrect type at runtime by throwing.

Includes minor fixups:
* Expose `RString` on the main thread
* Add `WebRDataScalar` as an abstract way to handle `RObject | string` for both `RMain.RObject` and `RWorker.RObject`.

### DIFF
--- a/src/webR/robj-worker.ts
+++ b/src/webR/robj-worker.ts
@@ -399,6 +399,9 @@ export class RSymbol extends RObject {
   // bad idea because this leaks memory.
   constructor(x: WebRDataScalar<string>) {
     if (x instanceof RObjectBase) {
+      if (Module._TYPEOF(x.ptr) !== RTypeMap.symbol) {
+        throw new Error('Unexpected object type in `RSymbol` constructor');
+      }
       super(x);
       return;
     }
@@ -450,6 +453,12 @@ export class RSymbol extends RObject {
 export class RPairlist extends RObject {
   constructor(val: WebRData) {
     if (val instanceof RObjectBase) {
+      if (
+        Module._TYPEOF(val.ptr) !== RTypeMap.pairlist &&
+        Module._TYPEOF(val.ptr) !== RTypeMap.call
+      ) {
+        throw new Error('Unexpected object type in `RPairlist` constructor');
+      }
       super(val);
       return this;
     }
@@ -555,6 +564,9 @@ export class RPairlist extends RObject {
 export class RList extends RObject {
   constructor(val: WebRData) {
     if (val instanceof RObjectBase) {
+      if (Module._TYPEOF(val.ptr) !== RTypeMap.list) {
+        throw new Error('Unexpected object type in `RList` constructor');
+      }
       super(val);
       return this;
     }
@@ -654,6 +666,9 @@ export class RString extends RObject {
   // Unlike symbols, strings are not cached and must thus be protected
   constructor(x: WebRDataScalar<string>) {
     if (x instanceof RObjectBase) {
+      if (Module._TYPEOF(x.ptr) !== RTypeMap.string) {
+        throw new Error('Unexpected object type in `RString` constructor');
+      }
       super(x);
       return;
     }
@@ -682,6 +697,9 @@ export class RString extends RObject {
 export class REnvironment extends RObject {
   constructor(val: WebRData = {}) {
     if (val instanceof RObjectBase) {
+      if (Module._TYPEOF(val.ptr) !== RTypeMap.environment) {
+        throw new Error('Unexpected object type in `REnvironment` constructor');
+      }
       super(val);
       return this;
     }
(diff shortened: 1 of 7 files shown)


src/webR/robj-worker.ts
```diff
@@ -450,6 +453,12 @@ export class RSymbol extends RObject {
 export class RPairlist extends RObject {
   constructor(val: WebRData) {
     if (val instanceof RObjectBase) {
+      if (
+        Module._TYPEOF(val.ptr) !== RTypeMap.pairlist &&
+        Module._TYPEOF(val.ptr) !== RTypeMap.call
```
lionel-, line 458: I think we should also fail for calls? Or at least convert them to pairlist, but failure probably makes more sense.

src/webR/robj-worker.ts
```diff
@@ -397,12 +397,15 @@ export class RSymbol extends RObject {
   // Note that symbols don't need to be protected. This also means
   // that allocating symbols in loops with random data is probably a
   // bad idea because this leaks memory.
-  constructor(x: RObject | string) {
+  constructor(x: WebRDataScalar<string>) {
     if (x instanceof RObjectBase) {
+      if (Module._TYPEOF(x.ptr) !== RTypeMap.symbol) {
+        throw new Error('Unexpected object type in `RSymbol` constructor');
+      }
```
lionel-, line 404: Would it be possible to wrap this in a generic function?

I think not but I'd love to be proven wrong. I have a hard time wrapping my mind around the fact that you can't associate types with constant values as in C++, so I intuitively reach for a generic for cases that are not supported. In typescript, generics are not templates, so you can't generate different runtime code paths with static values associated to types. I'll write that 50 times on the blackboard :-).

So I guess the next best thing is to take an optional `RTypeNumber` in the constructor of `RObject` or `RObjectBase` and check if `ptr` has the expected type if supplied.

src/webR/robj-worker.ts
```diff
@@ -887,6 +905,9 @@ abstract class RVectorAtomic<T extends atomicType> extends RObject {
 
 export class RLogical extends RVectorAtomic<boolean> {
   constructor(val: WebRDataAtomic<boolean>) {
+    if (val instanceof RObjectBase && Module._TYPEOF(val.ptr) !== RTypeMap.logical) {
+      throw new Error('Unexpected object type in `RLogical` constructor');
+    }
     super(val, RTypeMap.logical, RLogical.#newSetter);
```
lionel-, line 911: We already supply the type here so it makes especially sense to check in the parent ctor.

lionel- (approved):

src/webR/robj-worker.ts
georgestagg: We currently have situations where we create `RPairlist` instances wrapping R objects of type call. IIUC a call is just a special type of pairlist?

If we don't accept objects of type call here, I think we'll have to create a new class for them. How do you feel about a new `RCall` class extending `RPairlist`?

lionel-: Yes but a function is also a special type of pairlist in some sense. I think we want to be stricter here. Calls and pairlists have different semantics, only the latter are literals under evaluation.

I thought `RCall` existed already. It seems like a good idea to add that class. We can also introduce `RCall.toPairlist()` and `RPairlist.toCall()` if needed by existing code.

> RCall class extending RPairlist

I think pairlists and calls are independent types. If we had to create a hierarchy for node-like types, we'd have environments, functions, pairlists, calls etc inheriting from `RNode`. But we shouldn't do this since this hierarchy is an implementation detail from the point of view of the R API.

georgestagg: Okay, that makes sense. I've made an attempt at putting together an `RCall` class in the latest commit in this branch. For the moment I've just duplicated some pairlist-like methods `car()`/`cdr()`/`setcar()`. I also added an `exec()` method (so as to match `RFunction`'s `exec()`) that runs `Rf_eval()`.

I think this ended up as quite a nice addition, the new `RCall` class constructor and `exec()` can be used to simplify some places where we were constructing calls manually using pairlists.

I'll take another look over this Monday morning to confirm everything looks OK before attempting to merge.

src/webR/robj-worker.ts
georgestagg: I think we can create a function to do this, since we're working with R types and `RTypeMap` rather than strict TypeScript types here.

The latest version of the branch has these tests extracted into a function that throws if the R object passed does not match the R type also passed to it. To assist with that, I've moved `type()` from `RObject` into `RObjectBase`. I'm not yet 100% convinced about that move, but I think it makes sense since any `RPtr` reference will point to an R object with an associated type anyway.

src/webR/robj-worker.ts
georgestagg: That's a good observation, thanks!

src/webR/robj-worker.ts
lionel-: I think it should be `RCall.eval()` rather than `exec()`. Whereas `exec()` creates a call and evaluates it, an already constructed call only needs to be evaluated.

georgestagg: Good point, thanks. I've changed it to `RCall.eval()`.

PR #140, opened 2023-01-25
Use `R_UnwindProtect` when evaluating R calls

georgestagg:
Branched from #139.

This ensures that the stack is only unwound as far as the calling
function, rather than all the way up to the top level, before throwing a
JS Error while evaluating R calls in the worker thread.

This fixes an issue where JS `finally` blocks would continue to try to
unprotect objects after R has already jumped to top level and reset the
protection counter.

### DIFF
--- a/packages/webr/src/webr.c
+++ b/packages/webr/src/webr.c
@@ -34,27 +34,27 @@ struct safe_eval_data {
 };
 
 SEXP ffi_safe_eval(SEXP call, SEXP env) {
-  struct safe_eval_data* data = malloc(sizeof(struct safe_eval_data));
-  data->call = call;
-  data->env = env;
+  SEXP data = PROTECT(Rf_allocVector(RAWSXP, sizeof(struct safe_eval_data)));
+
+  struct safe_eval_data* ctx = (struct safe_eval_data *) RAW(data);
+  ctx->call = call;
+  ctx->env = env;
 
   SEXP cont = PROTECT(R_MakeUnwindCont());
   SEXP res = R_UnwindProtect(safe_eval_body, data, safe_eval_cleanup, cont, cont);
-
-  UNPROTECT(1);
-  free(data);
   return res;
 }
 
 static
 SEXP safe_eval_body(void* data) {
-  struct safe_eval_data* ctx = (struct safe_eval_data *) data;
+  struct safe_eval_data* ctx = (struct safe_eval_data *) RAW(data);
   return Rf_eval(ctx->call, ctx->env);
 }
 
 static
 void safe_eval_cleanup(void* cdata, Rboolean jump) {
 #ifdef __EMSCRIPTEN__
+  UNPROTECT(2);
   if (jump) {
     EM_ASM({
       throw new globalThis.Module.webr.UnwindProtectException(
--- a/packages/webr/src/webr.c
+++ b/packages/webr/src/webr.c
@@ -34,12 +34,14 @@ struct safe_eval_data {
 };
 
 SEXP ffi_safe_eval(SEXP call, SEXP env) {
+  // Unprotected in the R_UnwindProtect cleanup callback
   SEXP data = PROTECT(Rf_allocVector(RAWSXP, sizeof(struct safe_eval_data)));
 
   struct safe_eval_data* ctx = (struct safe_eval_data *) RAW(data);
   ctx->call = call;
   ctx->env = env;
 
+  // Unprotected in the R_UnwindProtect cleanup callback
   SEXP cont = PROTECT(R_MakeUnwindCont());
   SEXP res = R_UnwindProtect(safe_eval_body, data, safe_eval_cleanup, cont, cont);
   return res;
--- a/packages/webr/src/webr.c
+++ b/packages/webr/src/webr.c
@@ -2,6 +2,7 @@
 
 #include <R.h>
 #include <Rinternals.h>
+#include "decl/webr-decl.h"
 
 #ifdef __EMSCRIPTEN__
 #include <emscripten.h>
@@ -22,6 +23,46 @@ SEXP ffi_eval_js(SEXP code) {
   char eval_script[BUFSIZE];
   snprintf(eval_script, BUFSIZE, eval_template, R_CHAR(STRING_ELT(code, 0)));
   return Rf_ScalarInteger(emscripten_run_script_int(eval_script));
+#else
+    Rf_error("Function must be running under Emscripten.");
+#endif
+}
+
+struct safe_eval_data {
+  SEXP call;
+  SEXP env;
+};
+
+SEXP ffi_safe_eval(SEXP call, SEXP env) {
+  struct safe_eval_data* data = malloc(sizeof(struct safe_eval_data));
+  data->call = call;
+  data->env = env;
+
+  SEXP cont = PROTECT(R_MakeUnwindCont());
+  SEXP res = R_UnwindProtect(safe_eval_body, data, safe_eval_cleanup, cont, cont);
+
+  UNPROTECT(1);
+  free(data);
+  return res;
+}
+
+static
+SEXP safe_eval_body(void* data) {
+  struct safe_eval_data* ctx = (struct safe_eval_data *) data;
+  return Rf_eval(ctx->call, ctx->env);
+}
+
+static
+void safe_eval_cleanup(void* cdata, Rboolean jump) {
+#ifdef __EMSCRIPTEN__
+  if (jump) {
+    EM_ASM({
+      throw new globalThis.Module.webr.UnwindProtectException(
+        'A non-local transfer of control occured during evaluation',
+        $0
+      );
+    }, cdata);
+  }
 #else
     Rf_error("Function must be running under Emscripten.");
 #endif
--- a/src/webR/utils-r.ts
+++ b/src/webR/utils-r.ts
@@ -1,4 +1,4 @@
-import { Module, DictEmPtrs, dictEmFree } from './emscripten';
+import { Module, DictEmPtrs, dictEmFree, EmPtr } from './emscripten';
 import { WebRData, RPtr } from './robj';
 import { RObject, REnvironment, RHandle, handlePtr } from './robj-worker';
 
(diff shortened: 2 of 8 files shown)


src/webR/utils-r.ts
```diff
@@ -61,3 +61,42 @@ export function parseEvalBare(code: string, env: WebRData): RObject {
...
+let evalErrorMessage: string | undefined;
+
+function evalCallBody(data: EmPtr): RPtr {
+  const call = Module.getValue(data, 'i32');
+  const env = Module.getValue(data + 4, 'i32');
+  return Module._Rf_eval(call, env);
+}
+
+function evalCallHandler(cond: RPtr, _: EmPtr) {
+  evalErrorMessage = RObject.wrap(cond).get(1).toString();
```
georgestagg, line 76: I tried directly throwing the error here, but the protection stack becomes unbalanced. I guess the unwinding is not actually completed until `R_tryCatchError()` has returned so we should wait until then before throwing.

src/webR/utils-r.ts
```diff
@@ -61,3 +61,42 @@ export function parseEvalBare(code: string, env: WebRData): RObject {
...
+  }
+
+  const dataPtr = Module._malloc(8);
+
+  Module.setValue(dataPtr, handlePtr(call), '*');
+  Module.setValue(dataPtr + 4, handlePtr(env), '*');
+
+  // In case of an error ensure that the stack is unwound only as far as this
+  // function, rather than up to top level, before throwing a JS error.
+  const res = Module._R_tryCatchError(evalBodyPtr, dataPtr, evalHandlerPtr, 0);
```
lionel-, line 95: The stack might be unwound with other events than errors, for instance a restart invokation or a `Q` event in `browser()`.

It would be safer to use a general purpose longjump catcher like `R_UnwindProtect()`. Unfortunately there's no way to extract errors from longjump tokens, so we'd still need the trycatch to convert R errors to JS errors.

I guess the simplest way is to call `R_tryCatchError()` from an `R_ExecWithCleanup()` context to detect other kinds of longjumps and throw a generic error in that case.

The more accurate way would be to use `R_UnwindProtect()`, throw a special javascript exception that wraps the longjump token in case of early exit, and then resumes the jump from `readConsole()`. Then we don't need to capture R errors I think. This would be the job of `captureR()` as before.

The problem with this approach is that it's incompatible with catch-all blocks though. The longjump exception should not be caught on the way to `readConsole()`.

lionel- (commented):
Tricky! I think this goes in the right direction but we might be able to do better with `R_UnwindProtect()`, cf comment inline.

src/webR/utils-r.ts
```diff
@@ -61,3 +61,42 @@ export function parseEvalBare(code: string, env: WebRData): RObject {
     unprotect(prot.n);
   }
 }
+
+let evalBodyPtr: EmPtr | undefined;
+let evalHandlerPtr: EmPtr | undefined;
+let evalErrorMessage: string | undefined;
+
+function evalCallBody(data: EmPtr): RPtr {
```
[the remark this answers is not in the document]
georgestagg: As part of changes moving to using `R_UnwindProtect()`, I've moved the body and cleanup functions into `packages/webr/src/webr.c`, written as C functions. So, there is now a function named `ffi_safe_eval` that evaluates a call with `R_UnwindProtect`. The function has not been given an R frontend in the package, but webR can call directly into it from JS using `Module.LDSO` once the `webr.so` shared library has been loaded.

One tricky aspect to this is throwing a custom JavaScript exception from the C code. Luckily, this is easier than it used to be because `EM_ASM` works in side modules now, but still involves a little bit of setup so that the custom error class `UnwindProtectException` is available for use. The solution I've used for the moment is to add the `UnwindProtectException` class as a property on the global object `Module.webr`, so that C can get to and throw an exception of that type.

src/webR/utils-r.ts
georgestagg: Thanks for this great suggestion, I had not properly understood `R_UnwindProtect()` until now but it really looks like the right tool for the job. I've reimplmented things on this branch to try and match your third suggestion - using `R_UnwindProtect()` with a custom javascript exception.

As a quick confirmation of how I understand what should be happening:
When an error (or other stack unwinding event) occurs during `safeEval()` a JS exception of type  `UnwindProtectException` is created. The custom exception takes a copy of the unwind protect token and is then thrown. The exception is passed up the call stack as it unwinds up to `readConsole()`, somtimes the exception is caught and re-thrown, where necessary. In `readConsole()` the exception is caught and `R_ContinueUnwind()` is invoked using the associated token, which then finally proceeds with the `NORET` `longjmp`.

With this change, R error conditions are no longer caught and converted into JS exceptions as part of `saveEval()`. Instead, they are written to `stderr` as normal. When evaluating an R call during a channel request, but not using `captureR` machinery, the associated promise on the main thread is rejected with a generic message[1] and any errors can be observed by consuming `stderr` messages from the output queue.

This seems to all work, but I appriciate another pair of eyes taking a look to make sure I have understood it all correctly and that the implemenation looks correct.

[1] For the moment the generic message has been set to "A non-local transfer of control occured during evaluation".

lionel-: Your description matches my expectations about how this works.

I'll take another look in the morning with a fresher set of eyes :)

packages/webr/src/webr.c
```diff
@@ -26,3 +27,43 @@ SEXP ffi_eval_js(SEXP code) {
...
 #endif
 }
+
+struct safe_eval_data {
+  SEXP call;
+  SEXP env;
+};
+
+SEXP ffi_safe_eval(SEXP call, SEXP env) {
+  struct safe_eval_data* data = malloc(sizeof(struct safe_eval_data));
```
lionel-, line 37: In general we allocate in raw vectors using `Rf_allocVector(RAWSXP, size)`, protect the resulting SEXP, and then dereference it with `RAW()` to get the pointer.

This way the memory is managed by R and we're safe when a longjump occurs.

For instance, it's possible (although unlikely) that `R_MakeUnwindCont()` fails and jumps, bypassing the `free()` call.

packages/webr/src/webr.c
```diff
@@ -26,3 +27,43 @@ SEXP ffi_eval_js(SEXP code) {
...
+  data->call = call;
+  data->env = env;
+
+  SEXP cont = PROTECT(R_MakeUnwindCont());
```
lionel-, line 41: So I'm trying to think through what happens to the `safe_eval_data` and `UnwindCont` memory when a longjump happens. Probably we need to take care of it manually before jumping since we can't count on R to do it.

I think a simple `UNPROTECT(2)` (2 is assuming the struct is allocated in a RAWSXP) in cleanup function should be enough.

src/webR/utils-r.ts
```diff
@@ -61,3 +61,19 @@ export function parseEvalBare(code: string, env: WebRData): RObject {
...
+  cont: RPtr;
+  constructor(message: string, cont: RPtr) {
+    super(message);
+    this.name = 'UnwindProtectException';
+    this.cont = cont;
+  }
+}
+
+export function safeEval(call: RHandle, env: RHandle): RPtr {
+  return Module.LDSO.loadedLibsByName['/usr/lib/R/library/webr/libs/webr.so'].module.ffi_safe_eval(
```
lionel-, line 75: Interesting

src/webR/webr-worker.ts
```diff
@@ -437,6 +445,7 @@ function captureR(code: string, env?: WebRPayloadPtr, options: CaptureROptions =
     );
     protectInc(call, prot);
 
+    // Call Rf_eval directly here so as to support the option withHandlers: false
```
lionel-, line 448: This line is from the `R_tryCatchError()` implementation right? So I think we can now call `safeEval()`?

lionel- (approved):
Looking great!

Just needs some final tweaks to memory protection inside the unwind-protect context.

packages/webr/src/webr.c
georgestagg: Thanks, I've tweaked `ffi_safe_eval` to remove `malloc/free` and use `PROTECT(RRf_allocVector(RAWSXP, size))` to let R handle it instead.

packages/webr/src/webr.c
georgestagg: I've moved the `UNPROTECT(2)` from `ffi_safe_eval` into `safe_eval_cleanup` (but outside the `if (jump)` branch) so that it should always run and clean up the protected token and `RAWSXP`, even on non-local transfer of control.

src/webR/utils-r.ts
georgestagg: I thought about creating a `webr::safe_eval` R function in the package and grabbing a handle to it, but I wasn't sure how errors would propogate through the resulting nested stack  of`Rf_eval` -> `R_UnwindProtect` -> `Rf_eval`. In the end this seemed like a neater way to execute that C function.

If doing it this way does cause any issues in the future we should be able to rewrite the `safeEval` function to use the alternative method above.

packages/webr/src/webr.c
```diff
@@ -26,3 +27,43 @@ SEXP ffi_eval_js(SEXP code) {
...
 #endif
 }
+
+struct safe_eval_data {
+  SEXP call;
+  SEXP env;
+};
+
+SEXP ffi_safe_eval(SEXP call, SEXP env) {
+  SEXP data = PROTECT(Rf_allocVector(RAWSXP, sizeof(struct safe_eval_data)));
```
lionel-, line 37: Needs a comment that this is unprotected in the cleanup callback.

src/webR/webr-worker.ts
```diff
@@ -536,7 +545,15 @@ function init(config: Required<WebROptions>) {
       if (!chan) {
         throw new Error("Can't read console input without a communication channel");
       }
-      return chan.inputOrDispatch();
+      try {
+        return chan.inputOrDispatch();
+      } catch (e) {
+        if (e instanceof UnwindProtectException) {
+          Module._R_ContinueUnwind(e.cont);
```
lionel-, line 552: Now I'm a bit nervous about this setup. If there is an intermediate context on the stack that is the target of the longjump, for instance if there's an R-level `tryCatch()` from which we call back into JS and then run some failing R code with `safeEval()`, then the continuation token will point to that intermediate frame.

However, at this point, we have unwound all of the stack. So the data in the continuation is dangling. This needs some more thinking.

lionel-:
So our main problem is that when a longjump occurs, R restores the protection stack _before_ jumping instead of after. This means that if JS finalizers are executed during the longjump, on the way to the target, we mess up the already updated state.

By using `unwindProtect()`, we only let R update the state up to the unwind-protect context. Then we fetch a jump token and promise to resume the jump at some point.

The problem is that we have no idea where the jump is. It could possibly be somewhere in the middle of the JS stack. So I think to be 100% safe we have to abort the longjump and convert it to a simple JS error that will unwind to top level. Consequences:

- We'll still be able to capture errors by calling `tryCatch()` inside `safeEval()`, e.g. `captureR()` will still work as expected.
- Other kinds of jump will be forbidden. I can't think of any important use case though, so for a first approach it seems fine to do it this way.
- The main disadvantage, and a big one, is that when an error occurs we lose the context of that error (the message in the console). This will make it hard to debug.

Another way would be to modify R to update the state _after_ the jump. Then it'd be fine to call unprotect from finalizers. I'm not sure how feasible this is though.

lionel-:
> The main disadvantage, and a big one, is that when an error occurs we lose the context of that error (the message in the console). This will make it hard to debug.

Actually this is not true. When an error is not caught, the error message is printed before the jump, so the console will have the message. So I think this is the way to go. What do you think @georgestagg?

georgestagg:
A quick summary of our further discussion on Slack:

* An issue occurs when unwinding the JS stack all the way to `dispatch()` using our custom JS exception with continuation token. By the time we have unwound all the way to the top level, it is not possible to continue and jump if the target was actually somewhere on the stack (in principle, could happen with e.g. an R `tryCatch()` called some JS code that itself called `safeEval()`).

* The alternative solution of simply not `longjmp`-ing and instead throwing a general JS Error will also not work because if there exists some intermediate R calls on the stack between the JS portions of stack and the top level they will be unwound by our JS Error and lost. This will leave dangling R state even if we forced our own jump to top level in R.

* We think a solution is to _also_ capture continuation tokens in `evalJs()` and continue the jump from there. That way an inner `unwindProtect` unwinds up to the boundary of R and JS code, but no futher. If there is then a jump into an intermediate R stack it is valid to do so from here. Otherwise, the upstack `unwindProtect` captures the new `longjmp` and the JS unwinding continues.

* Eventually there will be no more `evalJs()` on the stack. At this point we assume there is no R code above us, the jump is to top level, and so unwind back to `dispatch()` and use the continuation token from there.

Downsides:
* We must be careful to use `evalJs()`, `webr::eval_js()`, or something similar if we ever want to call JS code from R.

lionel- (approved):

PR #141, opened 2023-02-08
Add types for channel messages

lionel-:
It is currently tricky to update a message type because the channel erases the types of messages. With this PR we now cast messages to specific types on both sides of the channel to ensure that they match.

The types are imported from a new file 'webr-chan'.

TODO: Add types for responses.

### DIFF
--- a/src/webR/proxy.ts
+++ b/src/webR/proxy.ts
@@ -100,7 +100,7 @@ function targetAsyncIterator(chan: ChannelMain, proxy: RProxy<RWorker.RObject>) 
 export function targetMethod(chan: ChannelMain, prop: string): any;
 export function targetMethod(chan: ChannelMain, prop: string, payload: WebRPayloadPtr): any;
 export function targetMethod(chan: ChannelMain, prop: string, payload?: WebRPayloadPtr): any {
-  return async (..._args: unknown[]) => {
+  return async (..._args: WebRData[]) => {
     const args = _args.map((arg) => {
       if (isRObject(arg)) {
         return arg._payload;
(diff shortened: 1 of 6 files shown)


src/webR/webr-chan.ts
```diff
@@ -0,0 +1,88 @@
+import { Message } from './chan/message';
+import { UUID as ShelterID } from './chan/task-common';
+import { WebRPayload, WebRPayloadPtr } from './payload';
+import { RType, WebRData } from './robj';
+
+export { isUUID as isShelterID, UUID as ShelterID } from './chan/task-common';
```
georgestagg, line 6: Neat!

src/webR/proxy.ts
```diff
@@ -108,10 +111,11 @@ export function targetMethod(chan: ChannelMain, prop: string, payload?: WebRPayl
       };
     });
 
-    const reply = await chan.request({
+    const msg: CallRObjectMethodMessage = {
       type: 'callRObjectMethod',
-      data: { payload, prop, args },
-    });
+      data: { payload, prop, args: args as WebRPayload[] },
```
georgestagg, line 116: It does not really matter, but if you'd like you can avoid the `args: args` here by casting types slightly earlier.

``` diff
       return {
         obj: replaceInObject(arg, isRObject, (obj: RObject) => obj._payload),
         payloadType: 'raw',
-      };
+      } as WebRPayload;
     });
 
     const msg: CallRObjectMethodMessage = {
       type: 'callRObjectMethod',
-      data: { payload, prop, args: args as WebRPayload[] },
+      data: { payload, prop, args },
     };
     const reply = await chan.request(msg);
```

georgestagg: It might also be worth doing,

``` diff
 export function targetMethod(chan: ChannelMain, prop: string, payload?: WebRPayloadPtr): any {
-  return async (..._args: unknown[]) => {
+  return async (..._args: WebRData[]) => {
     const args = _args.map((arg) => {
```

georgestagg (approved):
Looks great, thanks! I bet this change prevents a few bugs in future.

PR #142, opened 2023-02-08
Propagate payload errors from channel

lionel-:
Branched from #141.

- Automatically reject the promises of request-response messages when an error payload is returned.
- Transform channels in classes so that common methods may be inherited to reuse code.

### DIFF
--- a/src/webR/chan/channel-common.ts
+++ b/src/webR/chan/channel-common.ts
@@ -1,0 +1,62 @@
+import { SharedBufferChannelMain, SharedBufferChannelWorker } from './channel-shared';
+import { ServiceWorkerChannelMain, ServiceWorkerChannelWorker } from './channel-service';
+import { WebROptions } from '../webr-main';
+import { isCrossOrigin } from '../utils';
+import { IN_NODE } from '../compat';
+
+// This file refers to objects imported from `./channel-shared` and
+// `./channel-service.` These can't be included in `./channel` as this
+// causes a circular dependency issue.
+
+export const ChannelType = {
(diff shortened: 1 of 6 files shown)


src/webR/chan/channel-common.ts
```diff
@@ -0,0 +1,62 @@
+import { SharedBufferChannelMain, SharedBufferChannelWorker } from './channel-shared';
+import { ServiceWorkerChannelMain, ServiceWorkerChannelWorker } from './channel-service';
+import { WebROptions } from '../webr-main';
+import { isCrossOrigin } from '../utils';
+import { IN_NODE } from '../compat';
+
+// This file refers to objects imported from `./channel-shared` and
+// `./channel-service.` These can't be included in `./channel` as this
+// causes a circular dependency issue.
```
georgestagg, line 9: That's a shame, but I can't immediately see a way around it.

src/webR/chan/channel.ts
```diff
@@ -26,14 +23,59 @@ import { WebRPayload } from '../payload';
...
+  protected resolveResponse(msg: Response) {
+    const uuid = msg.data.uuid;
+    const handles = this.#parked.get(uuid);
+
+    if (handles) {
+      const payload = msg.data.resp as WebRPayload;
+      this.#parked.delete(uuid);
+
+      if (payload.payloadType === 'err') {
+        handles.reject(webRPayloadError(payload));
```
georgestagg, line 71: Very nice! This is much better than before.

georgestagg (approved):

PR #145, opened 2023-02-09
Simplify converting JS error to R error in evalJS

georgestagg:
Replaces converting a JS error into an R error by constructing an R `stop()` call, by just calling `Rf_error()` directly.

Uses Emscripten's `allocateUTF8OnStack()`, which I only just discovered, to avoid having to work around not being able `free` an allocated string after `Rf_error()` jumps.


lionel- (approved):
Nice!

PR #146, opened 2023-02-13
Add additional infrastructure for R later package

georgestagg:
We would like to be able to use the later R package as a prerequisite for webR compatibility with shiny, but for this to work the later package needs some support from webR. There are a few ways to implement later's delay mechanism in the context of a webR package,

1) JavaScript's `async/await` mechanism would probably work well, but until the [stack switching Wasm proposal](https://github.com/WebAssembly/stack-switching) is accepted and has been implemented in browsers, we are currently stuck blocking in the worker thread, unable to yield to the event loop.

2) We could spawn a new web worker, delay for some time in the new web worker, then signal back to the R worker thread that the delay time has passed. This would require careful expansion of the channel mechanism to co-ordinate messages between transient web workers and the R worker thread.

3) The later package currently implements a delay on POSIX platforms using pthreads. However, Emscripten pthreads are marked as experimental when used in combination with `SIDE_MODULE`s, and previous experiments with using the `-pthreads` flag when building R packages for wasm have not been successful.

So, as a first step, this PR implements a simpler form of solution 2). Rather than spinning up a new web worker and adding new code to co-ordinate messaging, the delay is instead just scheduled on the main thread using JSs `setTimeout()` function, and a message is sent back to the worker thread after the delay using the current channel mechanism.

* A Wasm function pointer can be invoked from the main thread using a new request type of `invokeWasmFunction`. The function can be invoked with an arbitrary number of arguments of type `number` and optionally return a `number`. Since pointers are just `number` this also allows for functions like `void* fn(void* data)`.

* The worker thread may output a new message of type `setTimeoutWasm`. The main thread has been set up to listen for this type of message. Rather than forwarding the message out to the application, the main thread will execute a `setTimeout` so that after a given delay an `invokeWasmFunction` request is then sent to the worker thread for execution.

See https://github.com/r-lib/later/commit/3e3f10f2f1bc69b9232fafe9a95a257a8f350b5a for an example of how a patched later package can be implemented using this new infrastructure.

### DIFF
--- a/src/webR/webr-main.ts
+++ b/src/webR/webr-main.ts
@@ -258,12 +258,12 @@ export class WebR {
      */
     while (msg.type === 'setTimeoutWasm') {
       setTimeout(
-        (ptr: EmPtr, data: EmPtr) => {
-          this.invokeWasmFunction(ptr, data);
+        (ptr: EmPtr, args: number[]) => {
+          this.invokeWasmFunction(ptr, ...args);
         },
         msg.data.delay as number,
         msg.data.ptr,
-        msg.data.data
+        msg.data.args
       );
       msg = await this.#chan.read();
     }
@@ -381,10 +381,10 @@ export class WebR {
     }
   }
 
-  async invokeWasmFunction(ptr: EmPtr, data: EmPtr): Promise<EmPtr> {
+  async invokeWasmFunction(ptr: EmPtr, ...args: number[]): Promise<EmPtr> {
     const msg = {
       type: 'invokeWasmFunction',
-      data: { ptr, data },
+      data: { ptr, args },
     } as InvokeWasmFunctionMessage;
     const resp = await this.#chan.request(msg);
     return resp.obj as EmPtr;
--- a/src/webR/webr-main.ts
+++ b/src/webR/webr-main.ts
@@ -360,6 +360,15 @@ export class WebR {
     }
   }
 
+  async invokeWasmFunction(ptr: EmPtr, data: EmPtr): Promise<EmPtr> {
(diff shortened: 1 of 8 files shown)


src/webR/webr-main.ts
```diff
@@ -247,7 +249,26 @@ export class WebR {
...
   async read(): Promise<Message> {
-    return await this.#chan.read();
+    let msg = await this.#chan.read();
+
+    /* Handle any output messages requesting a delayed invocation of a wasm
+     * function. Otherwise, return messages to the application.
+     * TODO: Reimplement without using the main thread once it is possible to
+     *       yield in the worker thread.
+     */
+    while (msg.type === 'setTimeoutWasm') {
```
lionel-, line 259: I think this was already the case before, but this makes it obvious that users need to pump incoming messages using `read()`, even if only to discard them (e.g. if they interact with R only through `eval()`).

lionel- (approved):
Looks elegant!

src/webR/webr-main.ts
georgestagg: That's a good point. It would be better if users only using `captureR()` didn't have to worry about reading from the main channel at all. I think it's unavoidable for the moment, especially when plotting. However, in the future when we are using `OffscreenCanvas` (or some other graphics device) to also capture plots, it could be feasible. 

Also, prompted by your comment I just realised that using `webR.flush()` bypasses this conditional 😬

I wonder if we could open our own side channel for internal messages like this instead, and manage our own `read()` loop for internal messaging. Will hold off merging and think about that...

PR #147, opened 2023-02-14
Remove handling of error payloads on the main side

lionel-:
Follow up for #142

This removes the notion of an error payload on the main side so that we don't have to handle the error case in switch statements.

### DIFF
--- a/src/webR/robj.ts
+++ b/src/webR/robj.ts
@@ -47,8 +47,10 @@ export type WebRDataRaw =
   | boolean
   | undefined
   | null
+  | void
   | Complex
   | Error
+  | Uint8Array
   | ArrayBuffer
   | ArrayBufferView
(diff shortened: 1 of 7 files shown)


src/webR/robj.ts
```diff
@@ -47,8 +47,10 @@ export type WebRDataRaw =
   | boolean
   | undefined
   | null
+  | void
   | Complex
   | Error
+  | Uint8Array
```
georgestagg, line 53: I think `ArrayBufferView` should cover `Uint8Array` and friends, so this line can probably be removed.

georgestagg (approved):

src/webR/robj.ts
lionel-: hmm it was necessary at some point but I ran into weird type issues during dev.

Now removed.

PR #148, opened 2023-02-14
Add non-sheltered eval methods

### LINKED ISSUE #128: Memory management in webR

lionel-:
Branched from #147.

These variants return raw JS objects through the channel and thus do not need memory management on the main side.
This completes the set of memory management features. Closes #128.

### DIFF
--- a/src/webR/webr-worker.ts
+++ b/src/webR/webr-worker.ts
@@ -9,8 +9,9 @@ import { replaceInObject, throwUnreachable } from './utils';
 import { WebRPayloadRaw, WebRPayloadPtr, WebRPayloadWorker, isWebRPayloadPtr } from './payload';
 import { RObject, isRObject, REnvironment, RList, getRWorkerClass } from './robj-worker';
 import { RCharacter, RString, keep, destroy, purge, shelters } from './robj-worker';
+import { RLogical, RInteger, RDouble } from './robj-worker';
 import { RPtr, RType, RTypeMap, WebRData, WebRDataRaw } from './robj';
-import { protectInc, unprotect, parseEvalBare, UnwindProtectException, safeEval } from './utils-r';
+import { protect, protectInc, unprotect, parseEvalBare, UnwindProtectException, safeEval } from './utils-r';
 import { generateUUID } from './chan/task-common';
 
 import {
@@ -243,21 +244,65 @@ function dispatch(msg: Message): void {
 
           case 'evalRRaw': {
             const msg = reqMsg as EvalRMessageRaw;
-            evalR(msg.data.code, msg.data.env);
+            const result = evalR(msg.data.code, msg.data.env);
 
-            let out = undefined;
-            switch (msg.data.outputType) {
-              case 'void':
-                break;
-              default:
-                throw new Error('Unexpected output type in `evalRRaw().');
+            protect(result);
+
+            const throwType = () => {
+              throw new Error(`Can't convert object of type ${result.type()} to ${msg.data.outputType}.`);
+            };
+
+            try {
+              let out: WebRDataRaw = undefined;
+              switch (msg.data.outputType) {
+                case 'void':
+                  break;
+                case 'boolean':
+                  switch (result.type()) {
+                    case 'logical':
+                      out = (result as RLogical).toBoolean();
+                      break;
+                    default:
+                      throwType();
+                  }
+                  break;
+                case 'number':
+                  switch (result.type()) {
+                    case 'logical':
+                      out = (result as RLogical).toBoolean();
+                      out = +out;
+                      break;
+                    case 'integer':
+                      out = (result as RInteger).toNumber();
+                      break;
+                    case 'double':
+                      out = (result as RDouble).toNumber();
+                      break;
+                    default:
+                      throwType();
+                  }
+                  break;
+                case 'string':
+                  switch (result.type()) {
+                    case 'character':
+                      out = (result as RCharacter).toString();
+                      break;
+                    default:
+                      throwType();
+                  }
+                  break;
+                default:
+                  throw new Error('Unexpected output type in `evalRRaw().');
+              }
+
+              write({
+                obj: out,
+                payloadType: 'raw',
+              });
+              break;
+            } finally {
+              unprotect(1);
             }
-
-            write({
-              obj: out,
-              payloadType: 'raw',
-            });
-            break;
           }
 
           case 'newRObject': {
--- a/src/webR/webr-worker.ts
+++ b/src/webR/webr-worker.ts
@@ -18,6 +18,7 @@ import {
   CaptureRMessage,
   CaptureROptions,
   EvalRMessage,
+  EvalRMessageRaw,
   FSMessage,
   FSReadFileMessage,
   FSWriteFileMessage,
@@ -236,6 +237,25 @@ function dispatch(msg: Message): void {
                 methods: RObject.getMethods(result),
               },
               payloadType: 'ptr',
+            });
+            break;
+          }
+
+          case 'evalRRaw': {
+            const msg = reqMsg as EvalRMessageRaw;
+            evalR(msg.data.code, msg.data.env);
+
+            let out = undefined;
+            switch (msg.data.outputType) {
+              case 'void':
+                break;
+              default:
+                throw new Error('Unexpected output type in `evalRRaw().');
+            }
+
+            write({
+              obj: out,
+              payloadType: 'raw',
             });
             break;
           }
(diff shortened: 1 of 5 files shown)


src/webR/webr-worker.ts
```diff
@@ -240,6 +242,69 @@ function dispatch(msg: Message): void {
...
+                      break;
+                    default:
+                      throwType();
+                  }
+                  break;
+                case 'number':
+                  switch (result.type()) {
+                    case 'logical':
+                      out = (result as RLogical).toBoolean();
+                      out = +out;
```
georgestagg, line 273: I don't think it should affect behaviour, but you could do something like
``` javascript
out = Number((result as RLogical).toBoolean());
```

or

``` javascript
out = Number(out)
```

which might make the cast to `number` clearer than using `+`.

georgestagg (approved):
Looks brill, thanks!

I think it's okay for the moment, but we should keep an eye on the nested `switch`s under `case evalRRaw:` to make sure they don't end up unwieldy in the future if we add more type conversions.

src/webR/webr-worker.ts
lionel-: Good point, done

PR #149, opened 2023-02-15
Followup tweaks for #148

georgestagg:
Apologies for missing them in review.


lionel- (approved):
Thanks, by the way you can also add another review to a merged PR.

PR #152, opened 2023-03-02
Move `env` argument to a property of `options` in `evalR` functions

georgestagg:
`CaptureROptions` is changed so that the `env` argument becomes a property of that interface.

`evalR` is changed so that the second argument is the entire `options` object, rather than just `env`. The argument remains optional and, as before, should default to something sensible.

`new REnvironment()` is used on the worker thread to construct the R environment to evaluate within. When passing an existing `REnvironment` this should not change anything, but doing this enables the `env` property to be of type `WebRData` and an environment to be constructed on the fly from JavaScript objects,

`await webR.evalR('foo + bar', { env: { foo: 1234, bar: 1 } })`

I thought this was really neat, so I also added it as an example in the docs as well as updating references to arguments of `evalR()` and `captureR()`.

### DIFF
--- a/src/docs/evaluating.qmd
+++ b/src/docs/evaluating.qmd
@@ -22,10 +22,14 @@ to a reference to the result of the computation, given as an
 The `options` argument provides advanced control over how the R code is
 evaluated and captured. For example, this can be used to enable R's
 [autoprint](https://cran.r-project.org/doc/manuals/r-release/R-ints.html#Autoprinting)
-feature. The `env` property of the `options` argument can be set when you wish
+feature.
+
+The `env` property of the `options` argument can be set when you wish
 to evaluate the R code within an R environment, rather than the default global
-environment. This could be used, for example, to avoid variable name clashes
-when evaluating multiple unrelated objects.
+environment. This could be used, for example, to sandbox an evaluation in a
+local environment to avoid `<-` creating variables in the global environment.
+Providing an environment is also a convenient way to pass data from Javascript
+to the R process.
 
 If any error conditions are raised during evaluation, they will be re-thrown as
 JavaScript exceptions. Other conditions and stream output is written to the
--- a/src/docs/evaluating.qmd
+++ b/src/docs/evaluating.qmd
@@ -11,17 +11,21 @@ evaluate R code and retrieve the result of the computation. The method takes two
 arguments,
 
 -   `code` - The R code to evaluate.
--   `env` - Optional: The environment to evaluate in. Default:
-    [`RObject.globalEnv()`](api/js/classes/RWorker.RObject.md#globalenv).
-
-The `env` argument can be set when you wish to evaluate the R code within an R
-environment, rather than the default global environment. This could be used, for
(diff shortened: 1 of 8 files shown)


src/docs/evaluating.qmd
```diff
@@ -11,18 +11,22 @@ evaluate R code and retrieve the result of the computation. The method takes two
...
 [`WebR.evalR()`](api/js/classes/WebR.WebR.md#evalr) returns a promise resolving
 to a reference to the result of the computation, given as an
 [`RObject`](api/js/modules/RMain.md#robject) proxy.
 
+The `options` argument provides advanced control over how the R code is
+evaluated and captured. For example, this can be used to enable R's
+[autoprint](https://cran.r-project.org/doc/manuals/r-release/R-ints.html#Autoprinting)
+feature. The `env` property of the `options` argument can be set when you wish
+to evaluate the R code within an R environment, rather than the default global
+environment. This could be used, for example, to avoid variable name clashes
```
lionel-, line 27: I'm not sure what this last sentence means. Maybe:

> This could be used, for example, to sandbox an evaluation in a local environment to avoid `<-` creating variables in the global environment.

lionel-: It's also a convenient way of passing in data from Javascript to the R process.

src/webR/webr-main.ts
```diff
@@ -291,77 +292,68 @@ export class WebR {
...
-    code: string,
-    env?: REnvironment,
-    options: CaptureROptions = {}
-  ): Promise<{
+  async captureR(code: string, options: CaptureROptions = {}): Promise<{
     result: RObject;
     output: unknown[];
   }> {
-    return this.globalShelter.captureR(code, env, options);
+    return this.globalShelter.captureR(code, options);
```
lionel-, line 312: By the way I was thinking we should remove the possibility of using `captureR()` with the global shelter (or at least make it harder to do) since it's so easy to leak memory with the `destroy()` method of releasing objects.

src/webR/webr-main.ts
```diff
@@ -291,77 +292,68 @@ export class WebR {
...
    * the JavaScript console.
    *
    * See {@link captureR} for a version of this function that allows for
-   * configuration of advanced options and for returning outputs as part of the
-   * result.
+   * capturing and returning outputs as part of the result.
    *
    * @param {string} code The R code to evaluate.
-   * @param {RObject} [env] The R environment to evaluate within.
+   * @param {CaptureROptions} [options] Options for the execution environment.
```
lionel-, line 325: Should we rename to `EvalROptions` since `evalR()` will be the main user-facing tool to evaluate R code?

lionel- (approved):

src/docs/evaluating.qmd
georgestagg: Thanks, that's a much better phrasing. I've combined your comments and re-written to,

> This could be used, for example, to sandbox an evaluation in a
> local environment to avoid `<-` creating variables in the global environment.
> Providing an environment is also a convenient way to pass data from Javascript
> to the R process.

PR #155, opened 2023-03-06
Allow `captureR` only when using a webR shelter

georgestagg:
Remove `webR.captureR()` method and update the documentation to reflect the changes.

### DIFF
--- a/src/docs/evaluating.qmd
+++ b/src/docs/evaluating.qmd
@@ -15,9 +15,8 @@ arguments,
     [`EvalROptions`](api/js/interfaces/WebRChan.EvalROptions.md) object
     controlling how the code is evaluated.
 
-[`WebR.evalR()`](api/js/classes/WebR.WebR.md#evalr) returns a promise resolving
-to a reference to the result of the computation, given as an
-[`RObject`](api/js/modules/RMain.md#robject) proxy.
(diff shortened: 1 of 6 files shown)


src/docs/evaluating.qmd
```diff
@@ -58,17 +59,20 @@ result.
 
 ## Evaluating R code and capturing output with `captureR`
 
-The [`WebR.captureR()`](api/js/classes/WebR.WebR.md#capturer) method can be used
-to both evaluate R code and retrieve the result of the computation, and capture
-any stream output or conditions raised. The method takes two arguments,
+Once a webR [shelter](objects.qmd#shelter) has been created for the purposes of
```
lionel-, line 62: Maybe start with a small introduction of why a shelter is necessary in the case of `captureR()`? E.g.:

Unlike `evalR()` which only returns one object, `captureR()` returns a variable number of objects when R conditions are captured. This makes memory management of individual objects with the `destroy()` method unwieldy. For this reason, `captureR()` requires the shelter approach with which you destroy all sheltered objects at once.

lionel- (approved):

src/docs/evaluating.qmd
georgestagg: Thanks, I've added a variant of this paragraph to explain the reasoning.

PR #156, opened 2023-03-06
Expose `evalRRaw()` as part of the webR API, and add support for returning JS arrays.

georgestagg:
I've stuck with the `evalRRaw` method name, rather than `evalRAs`, as it seems a little clearer when reading to my eyes.

* `evalRRaw` is made non-private and exposed on the `webR` class.
* Typing is updated to reflect return types based on `outputType` argument.
* Some additional argument possibilities and `switch` branches are added to handle returning JS arrays.
* Docs are updated to describe `evalRRaw`.

### DIFF
--- a/src/docs/evaluating.qmd
+++ b/src/docs/evaluating.qmd
@@ -43,11 +43,11 @@ automatically manage returned R objects using a webR
 
 ### Returning JavaScript values when evaluating R code
 
-A selection of convenience methods are provided that return the result of
-evaluating R code as a promise resolving to a JavaScript object, rather than an
-`RObject` reference. The benefit is that since the returned object is of a
-native JavaScript type, it does not need to be memory managed like an `RObject`
-result.
+A selection of convenience methods are available that return the result of
+evaluating R code as a promise resolving to a JavaScript object, rather than
+an `RObject` reference. The benefit is that since the returned object is of a
+raw JavaScript type, it does not need to be memory managed like an `RObject`
+result would be.
 
 | Method                                                            | Returned object type |
 |-----------------------------------------------------|--------------------------|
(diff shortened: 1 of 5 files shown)


src/docs/evaluating.qmd
```diff
@@ -56,6 +56,28 @@ result.
...
+return of JavaScript `Array` objects.
+
+| `objectType` argument value    | Returned object type |
+|--------------------------------|----------------------|
+| `'boolean'`   | `boolean` |
+| `'number'`    | `number`  |
+| `'string'`    | `string`  |
+| `'boolean[]'` | Array of `boolean` values     |
+| `'number[]'`  | Array of `number` values      |
+| `'string[]'`  | Array of `string` values      |
```
lionel-, line 72: That's a nice way of requesting array outputs.

src/docs/evaluating.qmd
```diff
@@ -56,6 +56,28 @@ result.
...
+| `'void'`      | No return value |
+
+::: callout-warning
+The `evalRRaw()` method and its related convenience methods require that the
+result of the R code evaluation is a vector of type `logical`, `integer`,
+`double` or `character` and must not contain missing values.
```
lionel-, line 78: Should we add a bool to allow missing values as `null`? I guess that would make the typescript wrappers rather complicated.

lionel- (approved):

src/docs/evaluating.qmd
georgestagg: I thought about that, but I was also worried about doubling the number of TypeScript signatures for the function.

I wonder if we should just allow missing values as `null` when returning arrays, and keep the current behaviour for scalars. That would better match the already existing `.toArray()` behaviour, though the return type changes from e.g. `Promise<boolean[]>` to `Promise<(boolean | null)[]>` and users will need to check for `null` or use `!` to assert non-null.

lionel-: I thought about that too, but it might be slightly surprising. Maybe keep the types in a separate file that could be automatically generated if needed?

Or let's just wait and see if missing values are needed/requested.

lionel-: It's probably fine to pull in R objects temporarily for these more complex cases.

georgestagg: Okay, let's wait and see. If we need to grab an array with `null` in the meantime we can go through an `RObject` and destroy it after.

PR #160, opened 2023-03-07
Rename `toTree()` to `toJs()`

georgestagg:
The following is renamed:
 * `toTree()` -> `toJs()`
 * `toTreeOptions` -> `toJsOptions`
 * `WebRDataTree` -> `WebRDataJs`
 * `WebRDataTree[X]` => `WebRDataJs[X]`

This change also moves the content of `tree.ts` back into the `robj.ts` module. My reasoning is as follows:

 * `robj.ts` already contains type definitions for the related `WebRData` and `WebRDataAtomic` types.
 * Since creating `tree.ts`, `robj.ts` file has been split into 3, with `robj-main.ts` and `robj-worker.ts`. The `robj.ts` file itself is now a much smaller common module and can reasonably accommodate `WebRDataJs`.
 * Removing `Tree` simplifies the module structure a little, which I think helps presentation in the docs.

I'd like to maintain the history of `tree.ts` somehow, but I'm not sure how to do that when combining the content of one file into another with git.


lionel-:
I used this strategy before to preserve the history while merging two files. Note that this is destructive and will wipe out any uncommited files or changes so the working directory should be clean:

```sh
SRC=R/cnd.R
TARGET=R/cnd-error.R

git mv $SRC $TARGET
git commit -n -m 'Save '$TARGET' history'
SAVED_HASH=`git rev-parse HEAD`
git reset --hard HEAD~1

git mv $SRC _temporary.R
git commit -n -m 'Save '$SRC' history'

git merge $SAVED_HASH
git commit -a -n -m 'Split '$TARGET' from '$SRC' 1/3'

git mv _temporary.R $SRC
git commit -n -m 'Split '$TARGET' from '$SRC' 2/3'
```

lionel-:
oops sorry the snippet above is for _splitting_ a file.

georgestagg:
I found a post on Raymond Chen's MS blog that shows how two files can be combined into a third maintaining line-by-line history, but I don't want to rename `robj.ts`.

Perhaps the history of `tree.ts` is not so important in this case anyway. GitHub claims it only has four commits starting from the last time the data type was renamed, all authored by me.

PR #174, opened 2023-03-14
Rename WebROptions properties to camelCase

georgestagg:
`WEBR_URL` -> `baseUrl`. This reads a little better to my eyes than `webRUrl`, and the context is clear enough since the interface is named `WebROptions`.

`PKG_URL` -> `repoUrl`. Since technically this points to a repository.

`SW_URL` -> `serviceWorkerUrl`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,0 +1,12 @@
+# webR (development version)
+
+## Breaking changes
+ * Rename the properties of `WebROptions` so that they are all in camelCase, consistent with the rest of the webR TypeScript source. We have made the decision to release the above breaking change quickly while there are still a relatively low number of affected users.
+
+## Bug fixes
+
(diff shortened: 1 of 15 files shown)


lionel- (approved):
Now that the first release is out, all user facing changes should be logged in a NEWS or changelog file.

Is it me or does `baseUrl` read like "base eurl" instead of "base URL"? I guess there's no purely satisfying solution here.

georgestagg:
How about `indexUrl`? That's practically the same as Pyodide's `indexURL`.

lionel-:
I meant that I tend to read `baseUrl` as "base earl" in my head. This had me reconsider the comments of the airbnb style guide about consistently casing acronyms according to their first letter.

Issue #170, opened 2023-03-10
mgcv errors on demo page


noamross:
Hello! Excellent work with this project. When running the demo at https://webr.r-wasm.org/latest, I get the following error:

```
> webr::install("mgcv")
Downloading webR package: lattice
Downloading webR package: nlme
Downloading webR package: Matrix
Downloading webR package: mgcv
> mod <- mgcv::gam(hp ~ s(disp), data=mtcars)

fatal Fortran runtime error(./../../../../src/modules/lapack/dlapack.f:114618): Internal error: RUNTIME_CHECK(catKind.has_value()) failed at extrema.cpp(243)
Aborted()
```

georgestagg:
Thank you for this bug report.

To build webR we are using an early development version of LLVM flang to compile LAPACK Fortran source code included in the base R distribution into WebAssembly. This flang compiler is not yet stable, and we are aware of a few issues that affect our wasm build of LAPACK. We expect that over time the development of flang will mature and issues like this one will go away as we upgrade the version of LLVM that we are using.

I've merged a patch that works around this particular error, as it also solves a problem in an unrelated function in base R (details are in the linked PR above). Do note, however, that while the patch does also help with mgcv, in this case it is not sufficient. From my limited further testing of mgcv, there are still some numerical convergence issues, most likely caused by other problematic Fortran code that mgcv relies on.

So, I am going to leave this issue open for now and we can check how things stand in the future once when we have upgraded the LLVM branch or release that we're based on. To help when revisiting in future, the following is a short example of the current state of running mgcv code, demonstrating the convergence issues.


RUNS:
```
> mod <- mgcv::bam(hp ~ s(disp), data=mtcars)
```

FAILS:
```
> mod <- mgcv::gam(hp ~ s(disp), data=mtcars)
Error in magic(G$y, G$X, msp, G$S, G$off, L = G$L, lsp0 = G$lsp0, G$rank,  : 
  magic, the gcv/ubre optimizer, failed to converge after 400 iterations.
```

PR #178, opened 2023-03-20
Avoid `MAXLOC` in R's LAPACK distribution, fixing `chol()` when `pivot=TRUE`

georgestagg:
Last week I spent some time attempting to debug the cause of #170. This PR summaries my findings and describes a resulting fix for the previously broken R function `chol(..., pivot=TRUE)`.

The `RUNTIME_CHECK(catKind.has_value())` crash in `extrema.cpp` comes from the LLVM fortran runtime, in the intrinsic function `MAXLOC`. The functions `MINLOC`, `MAXVAL`, `MINVAL` and probably others are also affected by the same bug.

The affected runtime functions take in fortran arrays in the form of so-called descriptors, which can be thought of as a struct containing a pointer to the array along with some additional information. The process of taking a fortran array and creating a descriptor is known as [emboxing](https://flang.llvm.org/docs/FIRLangRef.html#fir-embox-fir-emboxop).

There seems to be at least two related LLVM flang issues. The first, in FIR generation, is that the type information for the emboxed array elements is dropped when invoking a function. Indeed by taking a look at the generated FIR for a `MAXLOC` call we can see that the type of the elements of the array, `?xf64`, is cast to `none` before the call to `_FortranAMaxlocDim`.

```
%235 = fir.embox %234 : (!fir.heap<i32>) -> !fir.box<!fir.heap<i32>>
fir.store %235 to %29 : !fir.ref<!fir.box<!fir.heap<i32>>>
%236 = fir.address_of(@_QQcl.24142151609af5b892f44f877601ca1f) : !fir.ref<!fir.char<1,43>>
%237 = fir.convert %29 : (!fir.ref<!fir.box<!fir.heap<i32>>>) -> !fir.ref<!fir.box<none>>
%238 = fir.convert %232 : (!fir.box<!fir.array<?xf64>>) -> !fir.box<none>
%239 = fir.convert %c4 : (index) -> i32
%240 = fir.convert %236 : (!fir.ref<!fir.char<1,43>>) -> !fir.ref<i8>
%241 = fir.convert %233 : (!fir.box<i1>) -> !fir.box<none>
%242 = fir.call @_FortranAMaxlocDim(%237, %238, %239, %c1_i32, %240, %c114633_i32, %241, %false) : (!fir.ref<!fir.box<none>>, !fir.box<none>, i32, i32, !fir.ref<i8>, i32, !fir.box<none>, i1) -> none
```

This issue can be patched by hand, but it turns out to be insufficient. A further issue is that at least in some cases the `fir.embox` operation outputs machine code that fails to allocate memory for and populate the additional information attached to the descriptor, which should contain the array's length, the `TYPE`, and the `KIND` of the elements of the emboxed array. The implmentation of `MAXLOC` (and others) try to use this additional information to decide how to proceed at runtime, leading to the crash.

It looks like these issues are already known and being worked on, with [relevant commits as recent as Jan 10](https://github.com/llvm/llvm-project/commit/dbee2030a5651d76b543f2b3275489c3e049df05) in the main LLVM repo. I tried cherry-picking into the `f18-llvm-project` repo a selection of commits over the last year or so working on the FIR emboxing and descriptor generation code, but without success. I think we will have to wait until the work in the `fir-dev` branch we are currently based on is fully upstreamed into LLVM so that we can then upgrade and pick up these fixes and improvements that way. I tried the recent LLVM 16 release without succcess, but perhaps the work will be done by the time LLVM 17 arrives.

In the meantime, this PR patches R's distribution of LAPACK so as to avoid using `MAXLOC`. LAPACK only seems to use a small subset of `MAXLOC`s features, and only for `DOUBLE PRECISION` arrays. So, we can avoid it by reimplementing a similar function ourselves. The change fixes the R `chol()` function, so I have also removed the patch that ignores the test of `chol()` in the base R package tests.

This change does help with #170, but there still seems to be some other numerical stability issues going on, so there is still further work to be done before that issue can be closed.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,8 @@
 
  * Improve compatibility when running webR under Node (#167 & #171).
 
+ * Fix `chol()` with `pivot=TRUE` by working around a Fortran library issue.
+
 # webR 0.1.0
 
 Initial Release
--- a/patches/R-4.1.3/emscripten-avoid-testing-issues.diff
+++ b/patches/R-4.1.3/emscripten-avoid-testing-issues.diff
@@ -1,22 +1,3 @@
-Index: R-4.1.3/src/library/base/man/chol.Rd
-===================================================================
---- R-4.1.3.orig/src/library/base/man/chol.Rd
-+++ R-4.1.3/src/library/base/man/chol.Rd
-@@ -114,6 +114,7 @@ qr(m)$rank # is 2, as it should be
- # chol() unlike qr() does not use a tolerance.
- try(chol(m))
- 
-+\dontrun{ # Throws a Fortran assertion error
- (Q <- chol(m, pivot = TRUE))
- ## we can use this by
- pivot <- attr(Q, "pivot")
-@@ -125,5 +126,6 @@ try(chol(m))  # fails
- (Q <- chol(m, pivot = TRUE)) # warning
- crossprod(Q)  # not equal to m
- }
-+}
- \keyword{algebra}
- \keyword{array}
 Index: R-4.1.3/src/library/base/man/stopifnot.Rd
 ===================================================================
 --- R-4.1.3.orig/src/library/base/man/stopifnot.Rd
--- a/patches/R-4.1.3/flang-avoid-maxloc.diff
+++ b/patches/R-4.1.3/flang-avoid-maxloc.diff
@@ -1,0 +1,102 @@
+Index: R-4.1.3/src/modules/lapack/dlapack.f
+===================================================================
+--- R-4.1.3.orig/src/modules/lapack/dlapack.f
++++ R-4.1.3/src/modules/lapack/dlapack.f
+@@ -114497,6 +114497,20 @@
+ *> \ingroup doubleOTHERcomputational
+ *
+ *  =====================================================================
++      INTEGER FUNCTION MAXLOCD( N, ARR )
++      INTEGER          I, N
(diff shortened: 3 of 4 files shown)


lionel- (commented):
I'm not sure about this. I fear that by patching up this fortran code we might just be creating further issues, possibly making it harder to diagnose flang problems? I encountered the mgcv issue when I was running CMD check on CRAN packages, and noticed the user reports here and on github, and thought this'd get fixed by itself as the flang devs make progress towards compiling lapack and blas and passing test suites.

I think it'd help to make good bug reports with small reprexes though, as this would help prioritise the issues that are blockers for some packages. WDYT?

lionel-:
> possibly making it harder to diagnose flang problems

But also possibly making it possible to _discover_ further flang problems. Feel free to merge if you think this is worth it.

georgestagg:
Thanks for the thoughts!

In principle, I agree with you that we shouldn't patch up R's Fortran code too much, and instead be patient for the llvm/flang devs to make progress. On the other hand, I am fairly confident in this particular patch. It is a relatively minor change to LAPACK and I think the benefits of fixing the `chol()` function outweigh the drawbacks.

And, like you say, by working around this particular issue we can find other (potentially unknown) issues to report back to the flang devs.

I will merge, but leave further mgcv Fortran investigation alone for now. I'll update the issue thread with some text explaining the current situation.

PR #187, opened 2023-04-03
Build Cairo graphics library and its prerequisites, statically linking with webR

georgestagg:
This PR tweaks `/libs/Makefile` so that it builds Cairo and its prerequisites for Wasm, which are then statically linked as part of the R Wasm build. This allows us to use the default bitmap graphics devices (`png()` etc) since they can use Cario for rendering graphics.

This is a prerequisite for Shiny compatibility. We could limit ourselves to SVG, but support in Shiny requires a non-ideal wrapper (issue https://github.com/rstudio/shiny/issues/2057), so existing apps would need to be modified.

Even without thinking about Shiny, I think this is a good thing to have. With this change, we can consider removing the patch that adds the HTML `canvas` graphics device, or at least moving it out of base R and into the webr support package. Instead, we could generate bitmaps fully in the worker thread and transfer them to the main thread. That has the potential of making plot capture with tools like `captureR()` much easier (no need to wait for `OffscreenCanvas` support in Safari, for example).

The downside is that plots rendered this way require TrueType fonts on the virtual file system. At the moment a suite of ~Roboto~ Noto fonts are downloaded from Google Fonts at build time and bundled with webR. There's a balance to be made here between Unicode coverage/font quality and file size.

## Download size comparison

| File | Without Cairo (v0.1.1) | Without Cairo (dev) | With Cairo & Noto fonts
| ------------- | ------------- | --- |---|
| `R.bin.data`  | 28.6MB  | 12.8MB | 17.0MB
| `R.bin.wasm`  | 12.7MB  | 5.5MB |  6.1MB

TODO:
- [x] Consider the choices in `fonts.conf` - do we need to support more Unicode glyphs? Should we have fallbacks?
- [x] Consider the consequences of building and including several more wasm libraries as part of webR.
- [x] Benchmark changes in initial webR download size with this change.
- [x] ~~Take a look at the `sed` patches in `libs/Makefile`~~ This has now been tidied up a lot.
- [x] Think about how this fits in with the lazy network filesystem proposal.
- [x] Check extra system prerequisites for building new libs - add to docker and `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ R's source code and supporting libraries are written in both C/C++ and Fortran. 
 If you are compiling webR using the default toolchain, ensure that you first install the following required prerequisites:
  * [Emscripten SDK](https://emscripten.org/docs/getting_started/downloads.html) (>=3.1.25)
  * cmake
+ * gperf
  * liblzma
  * libpcre2
  * node (>=16.0.0)
--- a/patches/R-4.1.3/fontconfig-typeof.diff
+++ b/patches/R-4.1.3/fontconfig-typeof.diff
@@ -1,0 +1,18 @@
+Index: R-4.1.3/src/library/grDevices/src/cairo/cairoFns.c
+===================================================================
+--- R-4.1.3.orig/src/library/grDevices/src/cairo/cairoFns.c
++++ R-4.1.3/src/library/grDevices/src/cairo/cairoFns.c
+@@ -1316,10 +1316,13 @@ PangoCairo_Text(double x, double y,
+    with freetype2/fontconfig (FT implies FC in Cairo).
+ */
+ 
++#undef TYPEOF
+ #include <cairo-ft.h>
+ #include <ft2build.h> // currently included by cairo-ft.h
+ #include FT_FREETYPE_H
+ #include <fontconfig/fontconfig.h>
++#undef TYPEOF
++#define TYPEOF(x)     ((x)->sxpinfo.type)
+ 
+ SEXP in_CairoFT(void) 
+ {
--- a/patches/cairo-1.14.12/null-surface.diff
+++ b/patches/cairo-1.14.12/null-surface.diff
@@ -1,0 +1,129 @@
+Index: cairo-1.14.12/src/cairo-analysis-surface.c
+===================================================================
+--- cairo-1.14.12.orig/src/cairo-analysis-surface.c
++++ cairo-1.14.12/src/cairo-analysis-surface.c
+@@ -830,56 +830,64 @@ static cairo_int_status_t
+ /* null surface type: a surface that does nothing (has no side effects, yay!) */
+ 
+ static cairo_int_status_t
+-_return_success (void)
++_return_success_paint (void *surface,
++		       cairo_operator_t op,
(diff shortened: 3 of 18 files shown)


patches/R-4.1.3/fontconfig-typeof.diff
```diff
@@ -0,0 +1,18 @@
+Index: R-4.1.3/src/library/grDevices/src/cairo/cairoFns.c
+===================================================================
+--- R-4.1.3.orig/src/library/grDevices/src/cairo/cairoFns.c
++++ R-4.1.3/src/library/grDevices/src/cairo/cairoFns.c
+@@ -1316,10 +1316,13 @@ PangoCairo_Text(double x, double y,
+    with freetype2/fontconfig (FT implies FC in Cairo).
+ */
+ 
++#undef TYPEOF
```
georgestagg, line 9: There seems to be some cross-talk here with R's `TYPEOF` macro for SEXP objects. I am unsure what it does not happen with normal builds.

patches/cairo-1.14.12/null-surface.diff
```diff
@@ -0,0 +1,129 @@
+Index: cairo-1.14.12/src/cairo-analysis-surface.c
+===================================================================
+--- cairo-1.14.12.orig/src/cairo-analysis-surface.c
++++ cairo-1.14.12/src/cairo-analysis-surface.c
+@@ -830,56 +830,64 @@ static cairo_int_status_t
+ /* null surface type: a surface that does nothing (has no side effects, yay!) */
+ 
+ static cairo_int_status_t
+-_return_success (void)
```
georgestagg, line 9: These patches to the Cairo source are due to the usual Wasm issue of "no function pointer casting allowed".

Rather than casting, multiple functions returning `CAIRO_STATUS_SUCCESS` are compiled with the correct function signature.

README.md
georgestagg, line 1: Other than adding `gperf` to the list of prerequisites, these changes are mostly just rewording and typographical tweaks.

georgestagg:
In latest version have rebased and switched from using Roboto Fonts to using Noto Fonts.

In addition to wider coverage in the base font, Noto fonts work well for our planned use of a network-backed virtual filesystem, since multiple Noto font files exist for different types of languages (see e.g. Noto Sans Arabic) and we can use fontconfig's fallback mechanism so that the extra font files are only accessed/downloaded if glyphs are missing from the base Noto Sans font.

lionel- (approved):
It's probably worth looking into making this a shared dynlib so that this stack isn't fetched if not needed (e.g. most pkgdown examples). The Emscripten tooling for this is not ideal though, AFAIR, because they prefer to statically link and then tree-shake.

georgestagg:
A nice turn of events is that although the Cairo build is static, R happens to use it to build its own shared lib at `/usr/lib/R/library/grDevices/libs/cairo.so` on the VFS anyway. So I think we should be able to lazily download that file in the same way as planned for font files.

Saying that, the Emscripten port of Freetype _is_ linked into the main binary now, so I agree that it is worth a look at generating more shared libs in future. We can hopefully capture the Fortran runtime library in that same work.

PR #190, opened 2023-04-11
Add R proxy object type predicate functions for TypeScript

georgestagg:
Up until now, we have been asserting the type of objects returned from `webR.evalR` using the TypeScript `as` keyword,

``` javascript
const foo = await webR.evalR('123') as RDouble;
```

This works OK when we know the returned object type at compile time.

For unknown R objects, the type must be detected at runtime. This can be done reasonably well with `RObject.type()` and `as`, but the type assertions do not feel very elegant,

``` javascript
const foo = await webR.evalR('some_unknown_var');
const fooType = await foo.type();

if (fooType === 'double') {
  const bar = foo as RDouble;
  console.log(bar.toNumber());
} else if (fooType === 'character') {
  const bar = foo as RCharacter;
  console.log(bar.toString());
}
```

This PR adds and exports more type predicate functions in `robj-main`. This simplifies things, as TypeScript can use them to perform type narrowing automatically. This removes the need to use the `as` keyword to assert the type and simplifies the code,

``` javascript
const obj = await webR.evalR('some_unknown_var');

if (isRDouble(obj)){
  console.log(obj.toNumber());
} else if (isRCharacter(obj)) {
  console.log(obj.toString());
}
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@
 * Issue an output message of type `'closed'` when the webR communication channel closes.
 * Build Cairo graphics library and its prerequisites for Wasm as part of the webR build process. This allows the default Cairo-based graphics devices in R, such as `png()`, `bmp()` and `svg()`, to work in webR.
 * Update webR's version of R to 4.3.0.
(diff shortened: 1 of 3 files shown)


lionel- (approved):
Nice!

Issue #185, opened 2023-04-02
Memory and file system for current sessions 


jeroen:
I have a few questions about running webR in a long running node webserver:
 
 - If I start multiple R istances by running `const session = new x.WebR()` for each incoming http request on the server, do these sessions interfere with each other? Or is each `WebR` instance independent?
 - How do I completely destroy and free such an R instance, including its Shelter and all objects? Does it happen automatically when the object that was instantiated from `new x.WebR()` goes out of scope? Can I manually destroy it to immediately release all memory?
 - Is the virtual file system shared between R instances?
 
To illustrate, below a simple NodeJS script, that runs the same code in 5 parallel R sessions:

```js
const x  = require("@r-wasm/webr");

for (i in [1,2,3,4,5]) {
  const session = new x.WebR();
  session.init().then(function(){
    console.log("Session started!")
    session.evalR(`pi <- pi*2; pi`).then(function(obj){
      obj.toArray().then(function(data){
        console.log(data)
      });
    });
  });
}
```

georgestagg:
So far, our main focus and goals with the development of webR have been in the context of running R locally within a web browser, rather than on a server. As such, starting and stopping many instances of webR has not yet been well tested by us. In the browser environment, our assumption has been that one (or a few at most) webR instances will be created early in an application's execution cycle and then used throughout, with tools such as R environments used to isolate execution where required.

Nevertheless, things should work in a node webserver. Though I am unsure of the performance penalty of starting and stopping many webR instances. My gut feeling is that it is high, probably too high to start a new instance for every HTTP request.

> do these sessions interfere with each other? Or is each WebR instance independent?

Each webR instance will start a new JavaScript worker thread wherein the actual R wasm binary is executed. As such, they are largely independent. They should not interfere with one another.

> How do I completely destroy and free such an R instance

[`WebR.close()`](https://docs.r-wasm.org/webr/latest/api/js/classes/WebR.WebR.html#close) will close the established communication channel and stop the webR worker thread. At this point, all memory allocated in the R wasm process itself should be reclaimed by the JavaScript garbage collector. The webR instance should not be used after calling `close()`, promises will not resolve.

[...]

PR #193, opened 2023-04-11
Issue a 'closed' message when the webR communication channel closes

georgestagg:
As I briefly mentioned as part of my response to #185, an output message when the webR communication channel closes could be useful when multiple webR instances are initialised and closed over the lifetime of a long-running app.

For example, the message can be used to indicate when to break out of an otherwise infinite async loop that has been set up to wait for and read messages from the output queue.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # webR (development version)
 
 * Improve accessibility of xterm.js in the webR REPL app by enabling `screenReaderMode`.
+* Issue an output message of type `'closed'` when the webR communication channel closes.
 
 # webR 0.1.1
 
--- a/src/webR/chan/channel.ts
+++ b/src/webR/chan/channel.ts
@@ -69,6 +69,10 @@ export abstract class ChannelMain {
     return promise as Promise<WebRPayload>;
   }
 
+  protected putClosedMessage(): void {
+    this.outputQueue.put({ type: 'closed' });
+  }
+
   protected resolveResponse(msg: Response) {
     const uuid = msg.data.uuid;
     const handles = this.#parked.get(uuid);
--- a/src/console/console.ts
+++ b/src/console/console.ts
@@ -148,7 +148,8 @@ export class Console {
    * This loop waits for output from webR and dispatches callbacks based on the
    * message recieved.
    *
-   * The promise returned by this asynchronous function never resolves.
+   * The promise returned by this asynchronous function resolves only once the
+   * webR communication channel has closed.
    */
   async #run() {
     for (;;) {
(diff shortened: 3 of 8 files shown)


NEWS.md
```diff
@@ -2,6 +2,8 @@
 
 # webR 0.1.1
 
+* Issue an output message of type `'closed'` when the webR communication channel closes.
```
lionel-, line 5: oh we didn't release 0.1.1 already? Or should we bump the news file?

I think it'd be helpful to use the same 4-components convention as in tidyverse development. After a release we bump the whole package to a dev version like `0.1.1.9000`. That helps figuring out where in the development cycle we are: in the middle of a release (with 3 components), or during the dev cycle (with 4 components).

See https://r-pkgs.org/lifecycle.html#sec-lifecycle-version-number-tidyverse

src/webR/chan/channel.ts
```diff
@@ -69,6 +69,10 @@ export abstract class ChannelMain {
     return promise as Promise<WebRPayload>;
   }
 
+  protected outputClosedMessage(): void {
```
lionel-, line 72: A verb would be a little nicer?

```suggestion
  protected putClosedMessage(): void {
```

lionel- (approved):

NEWS.md
georgestagg: Whoops! My mistake, we did release 0.1.1, this should have been bumped. I'll commit to `main` to bump the version and tidy up `NEWS.md` before merging this PR.

Good idea, going forward we should have a development version convention and bump the version after a release. I like the 4-component convention, but IIRC I think npm can be pretty picky about package versions following semver. It probably does not matter much on our own machines, but if we ever want to push development releases to npm we might hit that wall.

So, how do you feel about using the [semver pre-release modifier](https://semver.org/#spec-item-9) instead? After releasing `0.1.1` we'd bump to `0.1.2-dev`. A further development bump would be `0.1.2-dev.1`. That should give us valid semver that are greater than the current release, but lesser than any subsequent proper release (`0.1.2`, `0.2.0`). The downside is that we could end up with a jump of `0.1.2-dev` to `0.2.0` without any `0.1.2` release.

The other nice thing about this modifier is that npm packages asking for, say, "^1.0.0" [will usually ignore pre-release values, even though they are strictly greater](https://stackoverflow.com/questions/64916954/does-caret-in-semver-matches-pre-release-when-pre-release-version-is-the-lat), choosing to install `1.1.1` over `1.1.2-dev`.

lionel-: yup sounds good, it makes sense to adapt the practices to the ecosystem conventions.

georgestagg: Oh, this is a little tricker than I thought. I wanted to have the built-in `webr` R package also have the same version number, but `0.1.2-dev` is invalid for R packages. I guess I'll just set that back to `0.1.1` for now, since we haven't yet changed anything there, and we can revisit it later when we come to modifying the package itself.

Perhaps the best we can do is `0.1.2-dev` for the npm package and `0.1.1.9000` for the built-in R package.

PR #196, opened 2023-04-12
Reduce the download size of `R.bin.wasm` and `R.bin.data`

georgestagg:
This PR reduces the amount of data webR initially downloads by removing data that is not strictly required to launch R.

* The `parallel` package is removed (I think older builds of webR removed this in the past? I'm unsure if it's really useful here since we only have 1 worker thread).

* The `/usr/lib/R/lib` directory is removed. This directory only contains copies of wasm libraries that are also downloaded by Emscripten from the `dist` directory at start up, so they do not need to also be included in `R.bin.data`.

* Package `doc`, `help`, and `html` directories are removed. Without `system()` or an external reader, there is no useful way to view these help files within webR right now in any case.

The directories can probably be restored when our lazy-filesystem proposal is more developed. With that, the content can be downloaded on demand if it is really required.

In addition,

 * By setting `-fvisibility=hidden` when building `libFortranRuntime.a`, unused runtime symbols are not re-exported on `R.bin.wasm`, reducing the output filesize.

## Rough Results (before server compression)

| File  | Before | After | % Original |
| ---- | ------- | ---- | ---- |
| `R.bin.data` | 25.3MB | 12.8 MB | 51%
| `R.bin.wasm` | 12.8MB | 5.5MB | 43%
| Total (REPL app) | 40.2MB | 21.7MB | 54%


tools/flang/Makefile
```diff
@@ -47,6 +47,7 @@ RUNTIME_CFLAGS := $(RUNTIME_CFLAGS)
 RUNTIME_CFLAGS += -I$(BUILD)/include -I$(BUILD)/tools/flang/runtime
 RUNTIME_CFLAGS += -I$(SOURCE)/flang/include -I$(SOURCE)/llvm/include
 RUNTIME_CFLAGS += -DFLANG_LITTLE_ENDIAN -fPIC
+RUNTIME_CFLAGS += -fvisibility=hidden
```
lionel-, line 50: I'm not sure how this works. Does the runtime lib explicitly sets visibility attributes on the API entry points so that only the non-static internal functions end up hidden?

lionel- (approved):
Nice!

tools/flang/Makefile
georgestagg: I was surprised by this too! I could be wrong but I don't *think* the reason it works is because runtime lib is explicitly setting `visible`, but instead because we can link the entry point functions anyway, even though they're hidden, because we're statically linking.

I'll write up how I think it's working using an example using system `clang`, rather than `emcc`.

---

First, let's make a library. I am aiming to show that there's a subtle difference in behaviour of `__attribute__((visibility("hidden")))` when linking statically vs. dynamically. I'll also include a standard and static function to compare.

`lib.c`:
``` c
int foo() {
    return 0;
}
__attribute__((visibility("hidden"))) int bar() {
    return 1;
}
static int baz() {
    return 2;
}
```

```
$ clang -c lib.c -o lib.o           
$ clang lib.o -shared -o lib.so
$ ar -crs lib.a lib.o   
```

Now, let's inspect the symbol tables,

```
$ nm lib.a  
lib.a(lib.o):
0000000000000008 T _bar
0000000000000000 T _foo
0000000000000000 t ltmp0
0000000000000010 s ltmp1

$ nm lib.so
0000000000003fb0 t _bar
0000000000003fa8 T _foo
```

As exected, `baz` is not listed because it's static. But, `bar` appears even though it's `hidden`! And it's a different type of symbol in each. In `lib.a` it's global and in `lib.so` it's local.

Now, let's make a binary `main` that calls `foo` and `bar`,

`main.c`:
``` c
extern int foo();
extern int bar();
int main() {
    foo();
    bar();
    return 0;
}
```

Next we confirm the shared library fails to link, as expected because `bar` is hidden and so local to `lib.so`,
``` 
$ clang -o main main.o lib.so
Undefined symbols for architecture arm64:
  "_bar", referenced from:
      _main in main.o
```

However, if we do the same thing but statically link `lib.a`, it works.

``` 
$ clang -o main main.o lib.a
$ nm main  
0000000100000000 T __mh_execute_header
0000000100003fa8 t _bar
0000000100003fa0 T _foo
0000000100003f70 T _main
```

Linking succeeds, and in the final binary the symbol `_bar` (hidden) is defined as local, while the symbol `_foo` (standard) is global.

In our application, `R.bin.wasm` is `main` and `libFortranRuntime.a` is `lib.a`. By setting `-fvisibility=hidden` and statically linking, all the symbols from `libFortranRuntime.a` are still added to `R.bin.wasm` but defined as local instead of global.

Finally, Emscripten has an extra optimisation: When `SIDE_MODULE`, `MAIN_MODULE` or `LINKABLE` is set, local symbols are not exported. So the output `R.bin.wasm` is smaller.

---

I am actually unsure if the binary content is still there in `R.bin.js` and just unaccessible, but enabling Emscripten's DCE with `MAIN_MODULE=2` would remove it if so - though this has other implications that are hard to solve. I tried it and we shave off another MB or so, but then we would need to manually list everything that was DCE'd that we actually want to keep.

lionel-: IIUC the linker is in charge of dealing with visibility of symbols. However when a static library is created there is no linking involved, just a bunch of object files packaged together. The attributes are still encoded in the object file though, so that other tools can inspect them. I guess Emscripten uses this info to avoid exporting the symbol?

georgestagg: Yes, it looks like you're correct.

I took a look at Emscripten's source: [Emscripten is setting the `--export-dynamic` flag](https://github.com/emscripten-core/emscripten/commit/a022cc78974941b64aea60586994c3e32650b255) when calling `wasm-ld`. Then, from what I can tell, building a Wasm executable with this flag set [means symbols marked with `visibility=default` are exported](https://lld.llvm.org/WebAssembly.html#exports).

We changed the Fortran symbols to have `visibility=hidden` instead of `visibility=default`, so when `wasm-ld` is called during Emscripten's `emcc` process it reads the attributes from the object files (contained in the static library) and so the symbols are not exported.

Since we're not doing full DCE with `MAIN_MODULE=2` I think this means the bitcode is still in the final Wasm binary somewhere, we just can't invoke it directly from JS, and the big saving comes from not outputting a big list of library function names in ASCII as part of the final `.wasm` file.

georgestagg: That's a good question! I don't think packages need to be statically linked to the Fortran runtime, but the answer for why is complicated.

Wasm package libraries like `stats.so` are actually Emscripten `SIDE_MODULE`s and are compiled like this,
```
emcc -s SIDE_MODULE=1 -s WASM_BIGINT -Oz -s DECLARE_ASM_MODULE_EXPORTS=0 -shared -L/__w/webr/webr/wasm/lib -o stats.so [...] lminfl.o -L../../../../lib -lRlapack -L../../../../lib -lRblas
```

From this we can see that `stats.so` does **not** get statically linked to `libFortranRuntime.a`. [The Emscripten docs say that it won't even get statically linked to system libraries like libc](https://emscripten.org/docs/compiling/Dynamic-Linking.html#overview-of-dynamic-linking), even though `MAIN_MODULE`s do.

The `stats.so` `SIDE_MODULE` **is** linked to the `libRblas.so` and `libRlapack.so` `SIDE_MODULE`s, but they too don't statically link to the Fortran runtime.

```
emcc -s SIDE_MODULE=1 -s WASM_BIGINT -Oz -s DECLARE_ASM_MODULE_EXPORTS=0 -shared  -L/__w/webr/webr/wasm/lib -o libRblas.so blas.o  cmplxblas.o 
emcc -s SIDE_MODULE=1 -s WASM_BIGINT -Oz -s DECLARE_ASM_MODULE_EXPORTS=0 -shared  -L/__w/webr/webr/wasm/lib -o libRlapack.so dlamch.o dlapack.o cmplx.o   -L"../../../lib" -lRblas  
```

Now I can see that it looks really strange that `stats.so` still works at all. Emscripten's `dlopen()` dynamically links at runtime, so how can it see the Fortran runtime symbols in the `MAIN_MODULE` that have been hidden by this PR?

I hadn't given it much thought until now, but I think the reason is due to how the`libRlapack.so` and `libRblas.so` `SIDE_MODULE`s are being loaded differently to `stats.so`. They are not opened with `dlopen()` but instead linked at load time. i.e `-lRlapack` and `-lRblas` are included in the command line to build `R.bin.wasm`. From the Emscripten documentation,

>If you are doing load time dynamic linking then any symbols needed by the side modules specified on the command line will be kept alive automatically.

My interpretation of this is that some Fortran runtime symbols, though marked as `visibility=hidden`, are actually still exported after all because Emscripten has noticed they are required for use in `libRlapack.so` or `libRblas.so`.

I guess `stats.so` does not use any runtime symbols not also used by Lapack or BLAS. I further guess that this means that while it currently works OK, we might potentially have to explicitly add a list of extra Fortran runtime symbols to export in the future, to support exotic R packages. This is not a difficult tweak to make if is indeed required, you just add them to an `EXPORTED_FUNCTIONS` list when issuing the `emcc` command.

-----

I think if we made a `libFortranRuntime.so` `SIDE_MODULE` the entire file would be loaded at the initial startup anyway. Certainly `libRlapack.so` currently is, which depends on Fortran. So we'd end up in the same situation with a big initial download and still have to do something like this to manage the large number of unused exports.

PR #198, opened 2023-04-12
Update install.R - omitted warning

Polkas:
The warning call is not used correctly because of the `cat` call inside it.
Please compare `warning(cat("Hello World\n"))` with `warning("Hello World\n")`.
Additionally, I expanded the warning message to be more helpful, and this update is optional.


georgestagg (commented):
Thanks, good spot!

Please allow me to make a suggestion and keep the error message short and simple for now.

georgestagg (approved):

Issue #203, opened 2023-04-28
Copying Uint8Array from JS to R seems extremely slow


jeroen:
Benchmarking my data export tool, I found that the bottleneck is not R, but copying binary data from JavaScript to the virtual FS. The other way around (from R to JS) is 1000x faster, so maybe I am doing something wrong.

For example, lets create a `Uint8Array` of 10MB:

```js
const webr = require("@r-wasm/webr");
const session = new webr.WebR();
const buffer = new ArrayBuffer(10*1024*1024);
const array = new Uint8Array(buffer);
await session.init();
```

Writing this to the R virtual filesystem takes 8 seconds:

```js
await session.FS.writeFile('test.bin', array);
```

Trying to copy the data into an R raw vector even takes 40 seconds:

```js
const d = await new session.RRaw(array);
```

Am I doing something wrong? By comparison, allocating a 10MB vector in webr, writing this to the virtual filesystem, and then reading it in JavaScript, only 21ms so this seems not be the issue

```js
// this is super fast
await session.evalRVoid("writeBin(raw(10*1024*1024), 'test.bin')");
var outbuf = await session.FS.readFile('test.bin');
var res = Buffer.from(outbuf, 'binary');
```

Here is a full script to run in node:

```js
const webr = require("@r-wasm/webr");
const session = new webr.WebR();

async function benchmark(){
	const buffer = new ArrayBuffer(10*1024*1024);
	const array = new Uint8Array(buffer);
	const t0 = new Date();
	await session.init();
	console.log(`Copying ${array.length} bytes to FS`)
	const t1 = new Date();
	await session.FS.writeFile('test.bin', array);
	const t2 = new Date();
	var outbuf = await session.FS.readFile('test.bin');
	var res = Buffer.from(outbuf, 'binary');
	console.log(`Copied ${res.length} bytes back to js`)
	const t3 = new Date();
	console.log("Init: ", t1-t0, "ms");
	console.log("writeFile: ", t2-t1, "ms");
	console.log("readFile: ", t3-t2, "ms");
	session.close();
}

benchmark()

// Init:  873 ms
// writeFile:  9871 ms
// readFile:  15 ms
```

georgestagg:
Thanks for digging into this and reporting back!

[...]

PR #204, opened 2023-05-02
Use msgpack to encode & decode data for channel transfer

georgestagg:
The [@msgpack/msgpack](https://www.npmjs.com/package/@msgpack/msgpack/v/2.8.0) npm library is a JS implementation of the serialisation format described at https://msgpack.org. It is designed to be a fast and efficient binary serialisation scheme, and IMO is a better fit for the types of messages (i.e. containing possibly large raw data) that we send over the webR channels.

Using msgpack, rather than encoding messages with JSON, leads to a significant performance improvement when sending large messages. See, for example, [the benchmark given in #203](https://github.com/r-wasm/webr/issues/203#issuecomment-1531078857). In that test there is a greater than 10x speedup in transferring data from the main thread to the worker thread with this change.

---

Additionally, a minor change is made to `toWebRData` fixing an issue where `ArrayBuffer` is treated like an object rather than an array, generating a very large `names` array containing stringified indexes (e.g. `['1', '2', ..., '10485760']`). In my own quick benchmark, this fix takes the time for an `await webR.RRaw(object)` for a 10MB object down from around 10 seconds to 500 ms.

### DIFF
--- a/src/package-lock.json
+++ b/src/package-lock.json
@@ -1,14 +1,15 @@
 {
   "name": "@r-wasm/webr",
-  "version": "0.1.0-alpha.18",
+  "version": "0.1.2-dev",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@r-wasm/webr",
-      "version": "0.1.0-alpha.18",
+      "version": "0.1.2-dev",
       "license": "SEE LICENSE IN LICENCE.md",
       "dependencies": {
+        "@msgpack/msgpack": "^2.8.0",
         "@types/emscripten": "^1.39.6",
         "jquery": "^3.6.0",
         "jstree": "^3.3.12",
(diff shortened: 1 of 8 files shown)


jeroen:
Perhaps orthogonal to this, but maybe there could be a special case for array buffers, when no encoding is needed?

We have the same issue in the V8 package: when copying data between R to JS, it gets encoded as json, except for arraybuffers which are directly copied into R's "raw vectors" and vice versa: https://github.com/jeroen/V8/blob/dff09bc83c552f9c2b74738b51704cf0db47d724/R/V8.R#L188-L197

So this provides a fast low level mechanism to share large binary data between R and JavaScript.

georgestagg:
At the moment with webR the `writeFile` message requires other information in addition to the raw file data,
```
{
  type: 'writeFile';
  data: {
    path: string;
    data: ArrayBufferView;
    flags?: string;
  };
}
```
Creating a new R raw vector with `new RRaw(data)` leads to a similar message structure where the `type` property encodes a particular operation.

In the context of the webR communication channel, I don't think it's a good idea for us to assume that a raw `ArrayBuffer` sent without any extra information means any one particular operation (even something as obvious as "turn this into an R raw atomic vector") since we also use the channel for several other purposes. Though, saying that, I am willing to revisit the idea in the future if binary data transfer continues to cause performance problems.

Another option would be to think about how to send standard messages and then follow up with (or have a side channel for) extra raw `ArrayBuffer` data. That would probably feel similar to using the `transfer` argument with `PostMessage`. Again, though, such a setup would require keeping track of some extra state so that the threads know what is going on, making things tricky to get right.

I think this PR is a good start. I assume that since msgpack advertises efficiency, it copies the `Uint8Array` data into the binary output with good performance. Certainly, it will be a vast improvement over encoding each byte into JSON and then parsing at the other end. Also, right now at least one buffer copy will always be required as part of the messaging scheme that we are using in place of `PostMessage` in any case.

jeroen:
OK thanks. I'm going to 🚀 this into production and see if I run into any issues.

jeroen:
So far things look good. The data export tool is much faster and it also seems to have resolved a mysterious bug where data would sometimes get corrupted in `writeFile`.

PR #207, opened 2023-05-04
Fix building of Cairo libs

lionel-:
I needed these changes to get Cairo libs to build:

- Disable pthreads with `-sUSE_PTHREADS=0`
- Disable XLib

PR #208, opened 2023-05-04
Fixes for newer Emscripten

lionel-:
Branched from #207

- Explicitly export runtime methods. Presumably needed because of this change:

  > Symbols that were previously exported from native code, but only for internal reasons, are no longer exported on the Module object by default. For example, previously if you were using Module._malloc but not explicitly including _malloc in EXPORTED_FUNCTIONS, it might have been exported anyway due to internal use of malloc within the JS library code.

- `module` symbol in LDSO elements has been renamed to `exports`


R/Makefile
```diff
@@ -113,7 +113,7 @@ MAIN_LDFLAGS += -s ALLOW_MEMORY_GROWTH=1
 MAIN_LDFLAGS += -s STACK_SIZE=1MB
 MAIN_LDFLAGS += -s EXIT_RUNTIME=1
 MAIN_LDFLAGS += -s ERROR_ON_UNDEFINED_SYMBOLS=0
-MAIN_LDFLAGS += -s EXPORTED_RUNTIME_METHODS=['allocateUTF8','UTF8ToString','callMain']
+MAIN_LDFLAGS += -s EXPORTED_RUNTIME_METHODS=['allocateUTF8','UTF8ToString','callMain','FS','ENV','getValue','setValue','LDSO','allocateUTF8OnStack','addFunction']
```
georgestagg, line 116: Very minor and probably not worth dealing with right now, but it would be nice if there was a way to build this argument up so that we don't have to put a big list of method names on a single line.

This might be harder than it sounds, I've noticed before that if you add a space after a comma here it breaks the build with a strange error. Maybe some clever quoting can deal with that though.

src/webR/utils-r.ts
```diff
@@ -72,7 +72,16 @@ export class UnwindProtectException extends Error {
...
+  // Old Emscripten
+  let exports: unknown = Module.LDSO.loadedLibsByName['/usr/lib/R/library/webr/libs/webr.so'].module;
+
+  // New Emscripten
+  // https://github.com/emscripten-core/emscripten/commit/c3bc5b89f4a0f0ebf53e8357179da54945c67239
+  if (!exports) {
+    exports = Module.LDSO.loadedLibsByName['/usr/lib/R/library/webr/libs/webr.so'].exports;
+  }
+
+  return (exports as any).ffi_safe_eval(
```
georgestagg, line 84: A neater alternative might be,

``` javascript
export function safeEval(call: RHandle, env: RHandle): RPtr {
  return Module.getWasmTableEntry(Module.GOT.ffi_safe_eval.value)(
    handlePtr(call),
    handlePtr(env)
  );
}
```

if it also works on the latest Emscripten. I know you said there's some problem with `getWasmTableEntry`, so perhaps it's no good...

Emscripten's `Module.GOT` (Global Object Table, maybe?) property might also be considered private, I am not sure. It would also requires a little typing added to `Module` in `emscripten.ts` to keep TypeScript happy,

``` javascript
  GOT: {
    [key: string]: {required: boolean; value: number};
  }
```

georgestagg (approved):
Looks good! I tried it on my local Emscripten v3.1.29 and it works fine there too.

Now seems like a good time to upgrade our targeted version of Emscripten anyway. Part of the next webR update will be an upgrade of R from 4.1 to 4.3, and as such the relevant binary R packages will live in a different directory in our CRAN-like repo. So, we don't have to worry too much about package compatibility with the current webR 0.1.1.

Once merged I will upgrade my locally installed Emscripten and the CI docker image to the latest version.

Some other minor comments are below.

src/webR/utils-r.ts
lionel-: Much better, done! As mentioned on Slack I needed to include `getWasmTableEntry()` in `Module` via `EXPORTED_FUNCTIONS` which causes a warning. I documented this in the commit message.

georgestagg: Another flag to try might be `-sDEFAULT_LIBRARY_FUNCS_TO_INCLUDE=['$getWasmTableEntry']` (I am unsure if the leading `$` is or is not necessary) as an alternative to `EXPORTED_FUNCTIONS`.

lionel-: I tried that but this gives:

```
warning: undefined symbol: getWasmTableEntry
```

With the leading `$` the leading `g` gets swallowed.

```
warning: undefined symbol: etWasmTableEntry
```

It feels like it should be available for `EXPORTED_RUNTIME_METHODS`, that may be a bug in Emscripten.

georgestagg: Thanks for trying.

> It feels like it should be available for `EXPORTED_RUNTIME_METHODS`, that may be a bug in Emscripten.

Agreed. [Emscripten's `Changelog.md`](https://github.com/emscripten-core/emscripten/blob/29cb1f5ad50c5e2b4dd521e137ca0499bbd9efb3/ChangeLog.md?plain=1#L416) suggests it should work.

Okay, let's merge with the warning if the build otherwise works and I'll take a closer look once I've upgraded my local Emscripten installation. If we can't figure it out we'll open an issue on the Emscripten repo.

R/Makefile
lionel-: We need the trick described in the "Splitting Without Adding Whitespace" section of https://www.gnu.org/software/make/manual/html_node/Splitting-Lines.html

PR #211, opened 2023-05-05
Enable optimisation during webr package build

georgestagg:
Also add `-L$(WEBR_ROOT)/wasm/lib` to `LDFLAGS` in case we want to link to wasm libs in future.


packages/webr-vars.mk
```diff
@@ -1,7 +1,13 @@
 include ${WASM_TOOLS}/fortran.mk
 -include ~/.webr-vars.mk
 
-R_INCLUDES = -I${R_SOURCE}/build-stage1/include/ -I${R_SOURCE}/src/include
+WEBR_ROOT = $(abspath ..)
+WASM_OPT ?= -Oz
+WEBR_INCLUDES = -I$(R_SOURCE)/build-stage1/include -I$(R_SOURCE)/src/include
```
lionel-, line 6: `WEBR_HOST_INCLUDES`? Or should we use stage2 includes here instead?

lionel- (approved):

packages/webr-vars.mk
georgestagg: Since the R makefiles are tweaked to build these extra packages they are added in both stages, even though the webr package is practically useless for host R. The stage2 includes haven't been made yet when building stage1, so we can't use them here.

Probably the right thing to do is use an environment variable or something to set the right includes directory for each stage. I'll add that before merging.

PR #214, opened 2023-05-12
Preliminary support for geospatial stack

lionel-:
- Add recipes for sqlite3, proj, and gdal
- Enable C++ exceptions. This uses an experimental wasm feature to avoid the overhead of emscripten emulation.
- Reorganise build variables so that all projects use the same flags. The flags are exported and no longer need to be forwarded manually.
- Add `OPTIONAL_WASM_LIBS`. Used in `make clean` and `make all`. All libs should add themselves either to this variable or to `DEFALT_WASM_LIBS`.

With this I can build the rgdal package. It is not fully functional yet but I'd like to get these changes merged before tackling the rest.


lionel-:
This section is a bit confusing, they also say:

> To use the WebAssembly EH and setjmp-longjmp support together:
> em++ -fwasm-exceptions -sSUPPORT_LONGJMP=wasm test.cpp -o test.js

So that's what I've used to be sure.

georgestagg:
Looks like the issue might be in the Emscripten ports version of FreeType. I can see when building in the docker image that `/opt/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/libfreetype.a` contains a reference to the `emscripten_longjmp` symbol.

I'll branch from this PR and experiment with building our own version of FreeType to see if that avoids the issue.

georgestagg:
The issue with FreeType has been fixed since Emscripten v3.1.35 (https://github.com/emscripten-core/emscripten/issues/19001). So, I spent some time today updating the GHA docker image to Emscripten v3.1.37, so as to match the version currently installed on my own machine.

In addition, node v17 or later is required for Wasm exceptions, and the version of node installed by `emsdk` by default is older than that. So, we also need to set an environment variable in GHA CI so that Emscripten uses the newer version of node that's installed by the `setup-node` action. A PR covering that extra change is at #215, which (hopefully) now builds cleanly in CI.

georgestagg (approved):

lionel-:
Thanks for fixing the build!

PR #219, opened 2023-05-17
Use system message channel for logging to console in `evalR()`

### LINKED ISSUE #210: evalR only prints first line of output when running under node
I noticed that only the first message or output gets emitted. 

```js
async function test(){
	const session = new webr.WebR();
	await session.evalRVoid("message('test 1')")
	await session.evalRVoid("message('test 2')")
	await session.evalRVoid("message('test 3')")
	await session.close();
}
test()
// test 1
```

The same happens when it is all in a single call, for example when installing two packages:

```js
async function test(){
	const session = new webr.WebR();
	await session.installPackages(['jsonlite', 'Matrix'])
}
test()
// Downloading webR package: jsonlite
```

Also for warnings:

```js
async function test(){
	const session = new webr.WebR();
	await session.evalRVoid('warning("warn1")')
	await session.evalRVoid('warning("warn2")')
	await session.close()
}
test()
// warn 1
```

georgestagg:
At the time of writing, logging to console from Worker threads does not work in Node if the worker (or main) thread is blocking -- yielding to the event loop is required. See https://github.com/nodejs/node/issues/30491 for further info.

This PR tweaks `evalR()` so that rather than directly logging to console from the worker thread, messages are instead sent to the main thread over the existing side-channel for internal system messages (currently only used for scheduling timeouts for Wasm functions).

Three new system message types are added: `console.log`, `console.warn` and `console.error`, which simply pass the message data to the associated logging function running on the main thread, rather than the worker thread.

This avoids the node logging issue, and so fixes #210. It also has a nice side effect of simplifying some test code that previously relied on a hack to spy on console output when running under Node.

### DIFF
--- a/src/tests/webR/webr-worker.test.ts
+++ b/src/tests/webR/webr-worker.test.ts
@@ -12,15 +12,19 @@ beforeAll(async () => {
 
 describe('Download and install binary webR packages', () => {
   test('Install packages via evalR', async () => {
+    const warnSpy = jest.spyOn(console, 'warn').mockImplementation((...args) => {});
     await webR.evalR('webr::install("cli", repos="https://repo.r-wasm.org/")');
     const pkg = (await webR.evalR('"cli" %in% library(cli)')) as RLogical;
     expect(await pkg.toBoolean()).toEqual(true);
+    warnSpy.mockRestore();
   });
 
   test('Install packages via API', async () => {
+    const warnSpy = jest.spyOn(console, 'warn').mockImplementation((...args) => {});
     await webR.installPackages(['MASS']);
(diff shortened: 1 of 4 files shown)


src/tests/webR/webr-worker.test.ts
```diff
@@ -12,15 +12,19 @@ beforeAll(async () => {
 
 describe('Download and install binary webR packages', () => {
   test('Install packages via evalR', async () => {
+    const warnSpy = jest.spyOn(console, 'warn').mockImplementation((...args) => {});
```
georgestagg, line 15: These mocks have been added simply to silence the message that's generated when installing R packages with `webr::install()`. We previously didn't see any output here due to the node logging issue.

lionel- (approved):

PR #232, opened 2023-06-22
Add custom error classes to be thrown by webR

### LINKED ISSUE #206: Use typed exceptions for errors raised by R
As discussed, it would be nice to identify exceptions raised by webr in JavaScript, so we can catch and handle them. 

Perhaps we could use a type like `WebRError` for exceptions that were caught in the worker thread, when throwing them in the main thread.

https://github.com/r-wasm/webr/blob/407cfba5b885ba805f25081e13df018ad9f24516/src/webR/payload.ts#L36-L42

georgestagg:
Closes #206.

A new class `WebRError` is introduced, extending the built-in JS `Error` class. The new error class is then further extended for more specific errors `WebRWorkerError`, `WebRChannelError`, and `WebRPayloadError`.

WebR errors thrown on the main thread have been changed to use the new `WebRError` classes.

Errors thrown on the worker thread have not been changed, but they are caught internally and forwarded to the main thread in any case. Any errors forwarded in this way are rethrown on the main thread as an instance of `WebRWorkerError`.

If an error more specific than `Error` is thrown on the worker thread, that error name is also forwarded to the main thread. Otherwise, the default error name of `WebRWorkerError` is used so as to match the class of the error, as expected.

The new error classes are exposed on the main webR entrypoint so that end-users can easily grab the classes for testing with `instanceof`.

With this, users can catch webR errors specifically, distinguishing them from other JavaScript errors. Useful for handling things like R syntax errors differently. For example,

``` js
import { WebR, WebRError } from '@r-wasm/webr';
[ ... init webR ...]

try {
  const val = await webR.evalRNumber(someRCode);
  // Do something with val
} except (e) {
  if (e instanceof WebRError) {
    // some webR error when evaluating the R code
  } else {
    // some other JS error when working with val
  }
}
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,6 +18,8 @@
 
 * Add a `quiet` argument to `webr::install()` and `webR.installPackages()` APIs that allows for silencing webR package downloading messages.
 
+* Add custom `WebRError` classes so that errors from webR can be identified by testing thrown exceptions with `instanceof WebRError`.
+
 ## Breaking changes
 
  * Removed the legacy `console.mjs` build target. The `Console` class is re-exported on the default `WebR` entrypoint, and so the extra build target is not required. This is technically a breaking change, but the legacy entrypoint was never documented and so we believe the change has only minor effect.
--- a/src/tests/webR/error.test.ts
+++ b/src/tests/webR/error.test.ts
@@ -1,0 +1,39 @@
+import { WebR, WebRError, WebRWorkerError } from '../../webR/webr-main';
+
+const webR = new WebR({
+  baseUrl: '../dist/',
+  RArgs: ['--quiet'],
+});
+
+beforeAll(async () => {
+  await webR.init();
+});
+
+test('Errors from webR are instances of WebRError', async () => {
+  const problem = webR.evalRNumber('"abc"');
+  await expect(problem).rejects.toThrow(WebRError);
+  await expect(problem).rejects.toThrow("Can't convert object of type character to number");
+});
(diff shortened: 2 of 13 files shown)


lionel- (approved):

lionel-:
Are the type errors surfacing as payload errors? It might be nice to have a type error class eventually.

georgestagg:
> Are the type errors surfacing as payload errors?

I don't think so. Type errors thrown from the worker thread will be re-thrown as instances of `WebRWorkerError` on the main thread. This should be reasonably OK, though, because we also set the error's [`name`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name) property based on the original error's name property. So if the `WebRWorkerError` is not caught, the console error output should still contain the string `"TypeError"` somewhere.

The `TypeError`s that were previously thrown in the communication channel code are now thrown as `WebRChannelError`s instead. I think this is a better fit for those particular errors in any case.

> It might be nice to have a type error class eventually.

Yes, this sounds like a good improvement for the situation of type errors thrown on the worker thread. I'll avoid adding any further complexity to this PR for the moment, but this is definitely something that can be added in a future expansion.

P.S. I think `WebRPayloadError` will only be thrown if something goes seriously wrong when passing objects between threads. For example, if the main thread was expecting an R object reference and gets a raw object instead. I think if this actually happens the user can't do much about it and it is probably an internal issue for the webR developers to investigate.

PR #240, opened 2023-07-18
Reimplement the webR demo REPL as a React app

georgestagg:
The original webR REPL demo app was implemented as a minimal TS app, with a little jQuery included for displaying and managing the virtual file system tree. This PR contains a rather large changeset reimplementing the REPL app in React.

This is a more modern JS framework to work with and allows the app to be separated into components that will be easier to maintain and extend going forward. It also allows us to finally drop the jQuery requirement, and instead rely on an accessible React library for displaying the virtual filesystem tree.

Reimplementing the features present in the original webR REPL app for the React framework largely happens in commit ccccc97455c528cb8767660405b9b4fa366e6457.

A Codemirror Editor component has been added to the app in the remaining commits, allowing for the authoring of longer chunks of R code on the web. The implementation details here have been heavily inspired by the implementation of the same editor in the rstudio/shinylive repo.

The Codemirror editor has been hooked up as an R pager so that help commands such as `?rnorm` should now work in webR. For default packages the help files will be downloaded on demand through the lazy VFS mechanism.

The final few commits in the series are just tidying up and keeping the linter happy.


lionel- (approved):
I've only done a very high level review as that is not my area of expertise.

Do/should we have a bare bones REPL demo based on the `Console` class that is framework agnostic? The React app is nice but may obscure the core logic with framework-specific logic. Though I guess

src/repl/App.tsx
```diff
@@ -0,0 +1,136 @@
...
+  REnv: {
+    R_HOME: '/usr/lib/R',
+    FONTCONFIG_PATH: '/etc/fonts',
+    R_ENABLE_JIT: '0',
+    COLORTERM: 'truecolor',
+  },
+});
+(globalThis as any).webR = webR;
+
+export interface TerminalInterface {
```
lionel-, line 23: Do we need these `Interface` suffixes? And if we do, should it be `ITerminal` which seems more conventional?

src/repl/App.tsx
```diff
@@ -0,0 +1,136 @@
...
+  read: Readline['read'];
+  write: Readline['write'];
+}
+
+export interface FilesInterface {
+  refreshFilesystem: () => Promise<void>;
+  openFileInEditor: (name: string, path: string, readOnly: boolean) => Promise<void>;
+}
+
+const terminalInterface: TerminalInterface = {
```
lionel-, line 34: Generally: It seems strange to have "interface" in the name of an actual object?

```suggestion
const terminal: TerminalInterface = {
```

src/repl/App.tsx
lionel-: We do need to differentiate from the main React class/function, in this case `Terminal`. So we have a `Terminal` component that takes an object containing hooks/callbacks set up by the application.

Is this a common pattern in React APIs, with perhaps a standard naming convention?

georgestagg:
> Do/should we have a bare bones REPL demo based on the Console class that is framework agnostic?

There's a bare-bones vanilla JS example using the `Console` class [given directly in the docs](https://docs.r-wasm.org/webr/latest/examples.html#creating-an-interactive-webr-repl-console). We could bundle an additional app similar to that example in the `src` dir if you think it would also be useful to also have a copy there?

lionel-:
> There's a bare-bones vanilla JS example using the Console class [given directly in the docs](https://docs.r-wasm.org/webr/latest/examples.html#creating-an-interactive-webr-repl-console). We could bundle an additional app similar to that example in the src dir if you think it would also be useful to also have a copy there?

yup I think it's useful to have an "Examples" folder that people can browse to quickly get a sense of the API style/verbosity.

src/repl/App.tsx
```diff
@@ -0,0 +1,136 @@
...
+          "    deleteFile: ", if (delete.file) "true" else "false", ",",
+          "  },",
+          "});"
+        ))
+        return(invisible(NULL))
+      }
+    )
+  `);
+  await webR.evalRVoid('webr::global_prompt_install()', { withHandlers: false });
+  await webR.evalRVoid('options(device=webr::canvas)');
```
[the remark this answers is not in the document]
georgestagg: Yes, good idea.

Since the pager function as implemented here is framework agnostic and just sends a message to the main thread, I'll move it into the `webr` package. It could be named something like `webr::pager_install()` so as to match `global_prompt_install()`.

Similarly, setting up canvas graphics could be added as `webr::canvas_install()`.

Not every application will need or want to include all of these setup steps, some will require just a subset or none at all. For example, shinylive will probably want to leave the default graphics device alone. I think exporting various `*_install()` functions from the `webr` package is a good level of abstraction.

In the end we'd end up with something like,

``` javascript
await webR.evalRVoid(`
  webr::pager_install()
  webr::global_prompt_install()
  webr::canvas_install()
`);
```

src/repl/App.tsx
```diff
@@ -0,0 +1,136 @@
...
+        const context = canvas.getContext('2d');
+        const msgData = output.data as CanvasMessage['data'];
+        if (msgData.event === 'canvasImage') {
+          context!.drawImage(msgData.image, 0, 0);
+        } else if (msgData.event === 'canvasNewPage') {
+          context!.clearRect(0, 0, canvas.width, canvas.height);
+        }
+        break;
+      }
+      case 'pager': {
```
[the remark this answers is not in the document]
georgestagg: This one really depends on how the app is set up and wants to handle the pager message and display the file. I'm not sure abstracting this handler into an exported function would be very reusable.

For the moment I'll move this branch (and perhaps the others) into a non-exported function, it's better organised that way in any case. If it turns out that other apps find these patterns useful and are copying them, we can export the function later.

src/repl/App.tsx
georgestagg: An update for anyone following along in future. After a little discussion privately, we've decided to stick with using the `Interface` naming for now.

For info, the pattern of moving state outside of components and passing handlers down through component props is called "[lifting state up](https://react.dev/learn/sharing-state-between-components)". This is a form of that pattern.

PR #243, opened 2023-07-21
Documentation updates: TypeScript examples, `RObject` typing, and `canvas()`/Cairo plotting.

georgestagg:


packages/webr/R/canvas.R
```diff
@@ -3,9 +3,18 @@
 #' A graphics device that generates HTML canvas API calls and executes them on
 #' the worker thread using a JavaScript `OffscreenCanvas`. Once drawing has
 #' finished, a JavaScript `ImageBitmap` object is transmitted to the main webR
-#' thread for display. The resulting webR output messages are of type `Message`
-#' with the `type` property set to `'canvasImage'` and the `data` property set
-#' to an object containing the image data, `{ image: ImageBitmap }`.
+#' thread for display.
+#'
+#' The resulting webR output messages are of type `Message` with the `type`
```
lionel-, line 8: Should we point to the type definitions for these messages here?

src/docs/examples.qmd
```diff
@@ -104,12 +166,56 @@ let output = await result.toArray();
...
+### Testing and narrowing R object types
+
+Since functions such as [`WebR.evalR()`](api/js/classes/WebR.WebR.md#evalr) may run arbitrary R code, it is not always clear what type of R object will be returned. To help manage the ambiguity, webR provides [type predicate functions](api/js/modules/RMain.md#functions) for several types of `RObject`.
+
+These functions return boolean `true` if the argument's R object type matches, and so different actions can be taken at runtime depending on the type of a given R object. This is particularly useful for TypeScript users, where the compiler is able to use the functions to narrow the R object type automatically.
+
+In this example, an R numeric, character, or list could be returned randomly by the `sample` function. Type predicate functions are used to detect and handle each case.
+
+``` typescript
+import { isRDouble, isRCharacter, isRList } from '@r-wasm/webr';
```
lionel-, line 194: These type predicates are great for internal dispatch but do we also have coercers/ctors that return a specific type or fail trying? I.e. `obj = webR.evalR(...); asRNumber(obj)` would be equivalent to `webR.evalRNumber(...)`.

lionel- (approved):

src/docs/examples.qmd
georgestagg: Yes, you can use the `RObject` ctors for this. For example, with the following code

``` javascript
  const foo = await webR.evalR('123');
  console.log(await foo.toJs());

  const bar = await new webR.RDouble(foo);
  console.log(await bar.toNumber());

  const baz = await new webR.RString(foo);
  console.log(await baz.toString());
```

TypeScript deduces that `foo` is an `RObject`, `bar` is an `RDouble`, and `baz` is an `RString`. At runtime, constructing the `RDouble` works, but `RString` fails as expected.

```
> {type: 'double', names: null, values: Array(1)}
> 123
> payload.ts:39 Uncaught (in promise) Error: Unexpected object type "double" when expecting type "string"
```

This was a good observation/example. I'll add it somewhere in this section of the docs.

Issue #262, opened 2023-08-31
Shimming install.packages ?


jeroen:
It might be helpful for new users if `install.packages()` is shimmed using `webr::install()`. Right now, the first thing people may want to try in the demo results in an error message that is difficult to understand.

<img width="697" alt="Screenshot 2023-08-31 at 2 47 39 PM" src="https://github.com/r-wasm/webr/assets/216319/b3e49976-1249-4bf0-93d6-46209c4096b2">

There is a precedent in the RStudio IDE of shimming install.packages:

```r
> install.packages
function (...) 
.rs.callAs(name, hook, original, ...)
<environment: 0x7feac5a466a0>
```

ColinFay:
+1 on this

ColinFay:
Note, `{renv}` also shims `install.packages` : https://github.com/rstudio/renv/blob/main/R/shims.R#L4

ColinFay:
@georgestagg happy to try and draft a PR for this if you want this feature

georgestagg:
Sure, contributions are welcome and I think this sounds like a reasonable feature.

PR #286, opened 2023-10-03
feat: shim install.packages

ColinFay:
This is a PR to shim `install.packages`, as defined in #262. 

I think the best place to do this is in the Rprofile.site (that is already present so we build on top of it). I've set the shim to happen when `{utils}` is loaded so that the new env is attached after `{utils}`. 

FWIW, I think we could have a generic function in `{webr}` that does this: this would allow to shim more functions from base if necessary. 

Happy to add some changes to this PR if necessary. 

Screenshot 2023-10-03 at 16 41 55
Screenshot 2023-10-03 at 16 42 04


ColinFay:
@georgestagg I would love to improve this by creating a generic `shim_webr` in the `{webr}` package, but making changes to the `packages/webr` package doesn't seem to be taken into account when rebuilding with make.

Any chance you could help me to do that?

Thanks

georgestagg:
Thanks for taking a look at this, Colin!

Yes, I agree that handling this in the `webr` package would be a good place. We already have some functions in that package to install some hooks, so installing the new shim(s) in the same place makes sense.

You can rebuild the `webr` package by changing into the ‘R’ directory and running:

```
make rebuild-extra-pkgs
```

This should also update the `dist` directory with the changes.

ColinFay:
Thanks @georgestagg, that worked ! 

I've moved the code to `webr:::shim_webr` — I've chosen to not export the function as it feels like there should be no reason for the standard user to call it, but feel free to tell me if you prefer it to be exported.

packages/webr/R/shims.R
```diff
@@ -0,0 +1,8 @@
+#' shim base functions
+shim_webr <- function(){
+  .e <- new.env()
+  .e[["install.packages"]] <- webr::install
+  setHook(packageEvent("utils", "attach"), function(...) {
+    attach(.e, name = "webr_shims", warn.conflicts = FALSE)
+  })
+}
```
georgestagg, line 8: Can you add a newline to the end of this file please?

packages/webr/R/shims.R
```diff
@@ -0,0 +1,8 @@
+#' shim base functions
+shim_webr <- function(){
```
georgestagg, line 2:
```suggestion
shim_install <- function(){
```

Do you mind if we name the function to match other setup functions in the webR package, e.g. `pager_install()`, `canvas_install()`?

R/Makefile
```diff
@@ -239,6 +239,7 @@ Rprofile: $(BUILD)/state/R-$(R_VERSION)/r-stage2
 	mkdir -p "$(R_WASM)/lib/R/etc/"
 	echo "options(expressions=400)" > "$(R_WASM)/lib/R/etc/Rprofile.site"
 	echo "options(bitmapType='cairo')" >> "$(R_WASM)/lib/R/etc/Rprofile.site"
+	echo "webr:::shim_webr()" >> "$(R_WASM)/lib/R/etc/Rprofile.site"
```
georgestagg, line 242: Perhaps it would be better to avoid shimming in the `.Rprofile`, and instead make it opt-in? This would match other changes we can optionally make with the webR package, such as setting up the Canvas graphics device and pager.

The idea would be that the function you've defined in the webR package is exported and documented (roxygen comments are good, they eventually filter out to the webR documentation website). Then, if a developer would like to opt-in they can just run the new function from JS using `evalRVoid()` as part of their setup.

For the REPL app, a new line would need to be added in `/src/repl/App.tsx` on line 97:
```
  await webR.evalRVoid('webr::shim_install()');
```
since the shim is no longer installed by default.

georgestagg (approved):
Thanks, this looks like a good implementation of the shim to me!

I've suggested a minor change to the setup, so that it's opt-in. Details are below.

ColinFay:
@georgestagg I've addressed your suggestions in the latest commits. 

I haven't touched the website documentation though, as I'm not sure if/where there should be some :)

georgestagg:
Looks great, thanks again!

> I haven't touched the website documentation though, as I'm not sure if/where there should be some :)

It should be automatic. If all goes well in CI, the function description should hopefully appear at https://docs.r-wasm.org/webr/latest/api/r.html

ColinFay:
@georgestagg I suppose there are other base functions that could be shimed too? 

Happy to contribute to others, let me know how / if I can help, and how you'd like to proceed for listing / implementing :)

PR #290, opened 2023-10-06
Add Plotly example

bugzpodder:
This adds a Plotly.js example.

You can view the live example here: 
https://pyscript.com/@bugzpodder/fancy-wind/latest

### DIFF
--- a/src/docs/examples.qmd
+++ b/src/docs/examples.qmd
@@ -367,14 +367,17 @@ After loading, the resulting web page should present an interactive R console si
 </script>
 ```
 
-### Interactive charts with ggplot2 and plotly
+### Interactive charts with `ggplot2` and `plotly`
 
-With ggplotly() by Plotly, you can convert your ggplot2 figures into interactive ones powered by plotly.js.
+Using the [`plotly`](https://plotly.com/r/getting-started/) R package, `ggplot2` output can be converted into interactive figures powered by [plotly.js](https://plotly.com/javascript/).
 ``` html
 <html>
   <head>
     <title>Plotly Example</title>
-    <script src="https://cdn.plot.ly/plotly-2.26.0.min.js" charset="utf-8"></script>
+    <script
+      src="https://cdnjs.cloudflare.com/ajax/libs/plotly.js/2.26.2/plotly.min.js"
+      charset="utf-8"
+    ></script>
   </head>
   <body>
     <div>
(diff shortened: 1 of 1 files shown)


src/docs/examples.qmd
```diff
@@ -366,6 +366,55 @@ After loading, the resulting web page should present an interactive R console si
...
+  "',",
+  "})"
+))`);
+    const msgs = await webR.flush();
+    msgs.forEach(msg => {
+      if (msg.type === 'plotly') {
+          outElem.replaceChildren();
+          Plotly.newPlot('out', JSON.parse(msg.data), {});
+      }
+    });
```
georgestagg, line 412: I think this would be better handled using [`webR.evalRString`](https://docs.r-wasm.org/webr/latest/evaluating.html#returning-javascript-values-when-evaluating-r-code):

```suggestion
      const plotlyData = await webR.evalRString(`
library(plotly)
library(ggplot2)

p <- ggplot(mpg, aes(displ, hwy, colour = class)) +
  geom_point()

plotly_json(p, pretty = FALSE)
`);
      outElem.replaceChildren();
      Plotly.newPlot('out', JSON.parse(plotlyData ), {});
```

georgestagg (changes requested):
I like this example of using `ggplot2` and `plotly` with webR, but I think it can be better handled without sending a custom communication message.

If the reason for the example was less about Plotly and instead about documenting sending custom messages, I personally think it is better not to have an example of sending a custom message using `eval_js()` as I would consider it a little too delicate for general use.

Instead, we should design a function to be added to the `webr` R package that deals with the required string interpolation behind the scenes. With this, the `chan.write()` JS function would be an internal implementation detail that R programmer does not need to know about.

bugzpodder:
Thanks I committed your suggestion and verified it in the link.
I personally use the following function since I prefer to deal with plotly and canvas using the same way using webR.flush():
```
plotly_chart = function (p) {
  library(plotly)
  json <- plotly::plotly_json(p, pretty=FALSE)
  webr::eval_js(paste0(
    "chan.write({",
    "  type: 'webR/plotly',",
    "  data: '",
    gsub("'", "\\\\'", gsub("\\\\", "\\\\\\\\", json)),
    "\',",
    "})"
  ))
}
```

georgestagg (approved):

PR #293, opened 2023-10-09
Expose Emscripten's `FS.mount` on webR's R and JS API

### LINKED ISSUE #260: How do I preload the R packages in a node app?
Hey, 

Given the following express app: 

```javascript
const app = require('express')()
const { WebR } = require('webr');

(async () => {
  globalThis.webR = new WebR();
  await globalThis.webR.init();
  console.log("webR is ready");
  await webR.installPackages(['attempt'])
  app.listen(3000, '0.0.0.0', () => {
    console.log('http://localhost:3000')
  })
})();


app.get('/', async (req, res) => {
  let result = await globalThis.webR.evalR('tools::toTitleCase("hello from r!")');
  let output = await result.toJs();
  res.send(output.values)
});
```

Every time I relaunch the app, a package installation is performed: 

```bash
$ node index.js 
webR is ready
Downloading webR package: rlang

Downloading webR package: attempt

http://localhost:3000

$ node index.js
webR is ready
Downloading webR package: rlang

Downloading webR package: attempt

http://localhost:3000
```

It works but I feel like there should be a better way to do it, because : 
- I feel like there is no good reason for downloading the packages every time the app is launched 
- If I have a lot of packages, it would take a lot of time every time
- I can't deploy this on a service that has no access to internet 

Wishlist : 

I would love to have something that allows (in the spirit of `node_modules`) to pre-install the packages on my machine and then load them at runtime. 

```javascript
(async () => {
 webR = new WebR();
  await webR.init();
  await webR.loadPackage( path.join(__dirname, '/r-packages/attempt'))
})();
```

Random thoughts: 

I could do it myself by downloading the binaries and toying the `WebRFS` interface but : 
- I would have to do the dependency tree resolution myself, while webr::install() and install.packages() already have everything set up.
- It might be a bit complex to read the file/folder tree and replicate it on webr filesystem. 

Happy to discuss this idea :) 

Cheers

georgestagg:
In the browser, this allows images built using Emscripten's `file_packager` tool to be mounted on the webR virtual filesystem using `webR.FS.mount()` in JS, and `webr::mount()` in R. This makes use of the Emscripten `WORKERFS` filesystem object.

Additionally, host machine directory paths can be mounted onto the VFS, enabling local file access, when webR is running under Node. This makes use of the Emscripten `NODEFS` filesystem object.

The exposed API allows one to choose between `WORKERFS` or `NODEFS`. The other Emscripten filesystem objects are currently unimplemented. Note that the environment must be correct (e.g. not allowed to use `NODEFS` in the browser) and webR will enforce this.

See https://emscripten.org/docs/api_reference/Filesystem-API.html#FS.mount for more information.

Closes #260, in that it should be possible to set up a host directory containing a set of R packages, mount the directory somewhere when running under Node, and then finally add that mount point to R's `.libPaths()`.

(Also fixes an unrelated documentation issue in `global_prompt_install`. I was going to rebase this away but it does not matter much.)


ColinFay:
Super awesome

packages/webr/R/mount.R
```diff
@@ -0,0 +1,47 @@
...
+#' filesystem type can only be used when webR is running under Node.
+#' 
+#' @param mountpoint a character string giving the path to a directory to mount
+#'   onto in the Emscripten virtual filesystem.
+#' @param data_url a character string giving the URL to a `.data` filesystem
+#'   image to be downloaded and mounted.
+#' @param path a character string giving the path to a host directory to be
+#'   mounted.
+#' @param type a character string giving the type of Emscripten filesystem to be
+#'   mounted: "workerfs" or "nodefs".
```
lionel-, line 25: It seems that the API would be simpler if this were two different functions. In general we try to avoid mutually exclusive arguments like `path` and `data_url` here. This would simplify usage and the documentation.

packages/webr/src/mount.c
```diff
@@ -0,0 +1,121 @@
...
+#endif
+
+SEXP ffi_mount_workerfs(SEXP base_url, SEXP mountpoint) {
+#ifdef __EMSCRIPTEN__
+  if (!Rf_isString(base_url) || LENGTH(base_url) != 1) {
+    Rf_error("URL must be a character string.");
+  }
+  
+  if (STRING_ELT(base_url, 0) == NA_STRING){
+    Rf_error("URL must not be `NA`.");
```
lionel-, line 17:
```suggestion
    Rf_error("URL can't be `NA`.");
```

packages/webr/src/mount.c
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  if (!Rf_isString(mountpoint) || LENGTH(mountpoint) != 1) {
+    Rf_error("`mountpoint` must be a character string.");
+  }
+  
+  if (STRING_ELT(mountpoint, 0) == NA_STRING){
+    Rf_error("`mountpoint` must not be `NA`.");
```
lionel-, line 25:
```suggestion
    Rf_error("`mountpoint` can't be `NA`.");
```

packages/webr/src/mount.c
lionel-, line 17: Worth wrapping in `check_string()`?

packages/webr/src/mount.c
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  EM_ASM({
+    const baseUrl = UTF8ToString($0);
+    const mountpoint = UTF8ToString($1);
+    const dataResp = Module.downloadFileContent(`${baseUrl}.data`);
+    const metaResp = Module.downloadFileContent(`${baseUrl}.js.metadata`);
+    if (dataResp.status < 200 || dataResp.status >= 300 ||
```
lionel-, line 33: Create paragraphs to give some breathing room?

```suggestion

    if (dataResp.status < 200 || dataResp.status >= 300 ||
```

packages/webr/src/mount.c
```diff
@@ -0,0 +1,121 @@
...
+        metaResp.status < 200 || metaResp.status >= 300) {
+      const msg = Module.allocateUTF8OnStack(
+        'Unable to download Emscripten filesystem image. ' +
+        'See the JavaScript console for further details.'
+      );
+      Module._Rf_error(msg);
+    }
+    const blob = new Blob([dataResp.response]);
```
lionel-, line 41:
```suggestion

    const blob = new Blob([dataResp.response]);
```

packages/webr/src/mount.c
```diff
@@ -0,0 +1,121 @@
...
+    Rf_error("`mountpoint` must be a character string.");
+  }
+  
+  if (STRING_ELT(mountpoint, 0) == NA_STRING){
+    Rf_error("`mountpoint` must not be `NA`.");
+  }
+
+  EM_ASM({
+    // Stop if we're not able to use a NODEFS filesystem object
+    if(typeof IN_NODE === 'boolean' && IN_NODE === false) {
```
lionel-, line 74:
```suggestion
    if (typeof IN_NODE === 'boolean' && IN_NODE === false) {
```

georgestagg:
Thanks for the review!

>  I think the API would be a bit cleaner with two different functions though. 

I agree, but I'd also like to try and match the already existing Emscripten API if possible.

Though, now that I think about it, while they do have a single `FS.mount()` function for multiple filesystem types, they don't have the mutually exclusive arguments. Instead, there is an `opts` argument that changes properties depending on the context.

In our case, perhaps we could combine the arguments? They are semantically different (a URL and a directory path), but at least both strings. I guess the argument could be named something generic like `source` to cover all bases.

So, maybe
```r
mount <- function(mountpoint, source, type = "workerfs")
```

which would then be used as in,

```r
mount("/mnt1", "./some/host/directory", type = "nodefs")
mount("/mnt2", "http://example.com/some/url.data", type = "workerfs")
```

On the other hand, I suppose that's just pushing complexity onto the documentation of `source`.

I will think some more about this before merging.

Issue #261, opened 2023-08-24
same WebR cannot handle multiple Promises


bugzpodder:
When using serviceworkers WebR has issues processing multiple requests concurrently.  This is not an issue when crossOriginIsolated = true.  I had to use a [workaround](https://github.com/r-wasm/webr/issues/255#issuecomment-1691677649) to get multiple code blocks to execute properly using a mutex.

[demo](https://pyscript.com/@bugzpodder/divine-hat/latest) crossOriginIsolated = false
Two promises installing lattice/ggplot2 resolving in parallel and neither returns.  This is not an issue with crossOriginIsolated = true, [demo2](https://bugzpodder.pyscriptapps.com/divine-hat-copy/latest/)

georgestagg:
Great spot that the issue only occurs under the Service Worker channel! That's why I missed reproduction in #255, lesson learned.

I've tagged this as a confirmed bug, thanks for taking the time to build examples.

PR #297, opened 2023-10-16
Fallback to the `PostMessage` channel when using `ChannelType.Automatic`

georgestagg:
With this change, when using `ChannelType.Automatic` channel (the default), the `PostMessage` channel is used as the fallback when the web page is not Cross-Origin Isolated, replacing `ServiceWorker`.

The `PostMessage` channel has the widest compatibility, but is unable to block for input (e.g. `readline()`, `menu()`, `browser()`, etc). 

After the introduction of the `PostMessage` channel for use with Shinylive, the response has been positive from other users too, praising the wider compatibility with static web servers. After discussion with @wch and under [suggestion](https://github.com/coatless/quarto-webr/issues/31#issuecomment-1720961696) from @coatless, I think it makes sense for `PostMessage` to be the default fallback position for automatic mode:

1) The previous `ServiceWorker` fallback was never really "automatic", it required extra setup in terms of placing special files on the server.
2) Before or after the change, we have the situation where there is some kind of reduced functionality when the page is not COI, but adding COI headers enables the extra functionality. This seems reasonable to me.

Note that this is a breaking change. If blocking for input is required, and COI is not possible, the `ServiceWorker` channel is still available as an alternative, but the user code that loads webR must be changed to request it explicitly.

(P.S. If the `ServiceWorker` channel turns out to be rarely used after this change, we could consider deprecation. My impression is that the performance impact and other issues (#261) with this channel make it rarely useful, particularly when loading R packages.)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,10 @@
 
 * Expose Emscripten's `FS.mount()` on `webr::mount` in R, and `webR.FS.mount()` in JavaScript. This allows images built using Emscripten's `file_packager` to be mounted on the virtual filesystem and host directory paths to be mounted when running under Node.
 
+## Breaking changes
+
+* When starting webR using the `ChannelType.Automatic` channel (the default), the `PostMessage` channel is now used as the fallback when the web page is not Cross-Origin Isolated. The `PostMessage` channel has the widest compatibility, but is unable to use functions that block for input (e.g. `readline()`, `menu()`, `browser()`, etc). If blocking for input is required, the `ServiceWorker` channel is still available, but must be requested explicitly.
+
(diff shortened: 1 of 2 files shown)


coatless:
I strongly endorse replacing the usage of `ServiceWorker` with `PostMessage` within the `Automatic` routine. This change is critical for broader adoption, as many users prefer to stick with the default settings.

lionel- (approved):
If the service worker has performance and usability problems then I agree it would make sense to deprecate it. This would make it easier to document and understand webR setups: a cross-origin isolated setup that fully supports the R REPL, and a simpler setup that only supports running R requests.

PR #298, opened 2023-10-16
Raise R error conditions on `system()` and when unable to block for input.

### LINKED ISSUE #279: Raise R errors instead of silent return for unsupported functions
Some functions like `system()` and `readline()` (the latter only since disabling the COI header approach) are not supported by webR, and return default values (`c("", "")` and `""` respectively, for these examples). I think it'd be better to raise an R error that these are not supported.

georgestagg:
Includes minor tweaks to `Sys.which` and `sessionInfo` to avoid using `system()` when running under Emscripten.

Closes #279.

### DIFF
--- a/packages/webr/R/testing.R
+++ b/packages/webr/R/testing.R
@@ -64,6 +64,8 @@ remove_lines_in_file <- function(src_file, lines) {
 #' @param pkg Name of the package to test.
 #' @returns 0 if the test was successful, otherwise 1.
 test_package <- function(pkg) {
+  op <- options(webr.hook_system = function(...) c("", ""))
+  on.exit(options(op), add = TRUE)
   old_wd <- getwd()
   pkgdir <- find.package(pkg)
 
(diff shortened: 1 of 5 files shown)


packages/webr/R/testing.R
```diff
@@ -64,6 +64,8 @@ remove_lines_in_file <- function(src_file, lines) {
 #' @param pkg Name of the package to test.
 #' @returns 0 if the test was successful, otherwise 1.
 test_package <- function(pkg) {
+  op <- options(webr.hook_system = function(...) c("", ""))
+  on.exit(options(op), add = TRUE)
```
georgestagg, line 68: The old setup where `system()` returns `c("", "")` has been added back here so that self-testing the `base` package continues to pass. [One of the examples in `base`](https://github.com/wch/r-source/blob/18022ab25fb0b45445360fafe4a9ca618f6d52b3/src/library/base/man/Platform.Rd#L102) assumes that `system()` is available.

lionel- (approved):

PR #305, opened 2023-10-19
fix: allow to pass multiple repo to install()

### LINKED ISSUE #299: Bug in webr::install for length repos > 1
We’ve been doing some experimental building of wasm binaries in r-universe. There is a bug in `webr::install` when options(repos) is of length > 1. For example when we do:

```r
webr::install('dplyr', repos = c('https://tidyverse.r-universe.dev', 'https://repo.r-wasm.org'))
```

R does not install dependencies from the second repo.

ColinFay:
Will fix #299


jeroen:
Thanks! Looks good to me.

georgestagg:
Yep, that looks good. Thanks a lot!

PR #312, opened 2023-11-04
Docker: use R installation manager

jeroen:
Switch the host R installation to use [rig](https://github.com/r-lib/rig) from @gaborcsardi.

This makes it much easier to pin the host installation to match the webr version (4.3.0 currently). Also it automatically sets up pak and PPM binaries, so I think this will much simplify management.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,17 +18,11 @@ RUN mkdir -p /etc/apt/keyrings && \
     apt-get update && \
     apt-get install nodejs -y
 
-# Install standard R so that we can use PPM binaries
-RUN apt-get install -y --no-install-recommends software-properties-common && \
-    add-apt-repository --enable-source --yes "ppa:marutter/rrutter4.0" && \
-    add-apt-repository --enable-source --yes "ppa:c2d4u.team/c2d4u4.0+" && \
-    apt-get purge -y software-properties-common && \
-    apt-get autoremove -y
+RUN curl -L https://rig.r-pkg.org/deb/rig.gpg -o /etc/apt/trusted.gpg.d/rig.gpg &&\
+    sh -c 'echo "deb http://rig.r-pkg.org/deb rig main" > /etc/apt/sources.list.d/rig.list' &&\
+    apt update && apt install r-rig
 
-RUN apt-get update && apt-get install -y --no-install-recommends \
-        r-base r-base-dev r-recommended && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
+RUN rig add 4.3.0
 
 # Download webR and configure for LLVM flang
 RUN git clone --depth=1 https://github.com/r-wasm/webr.git /opt/webr
@@ -44,21 +38,6 @@ RUN R CMD INSTALL packages/webr
 # Build webR with supporting Wasm libs
 RUN cd libs && make all
 RUN make
-
-# Setup PPM for binary native R packages
-COPY <<EOF /root/.Rprofile
-options(repos = c(CRAN = "https://packagemanager.posit.co/cran/__linux__/jammy/latest"))
-options(HTTPUserAgent = sprintf(
-    "R/%s R (%s)",
-    getRversion(),
-    paste(
-        getRversion(),
-        R.version["platform"],
-        R.version["arch"],
-        R.version["os"]
-    )
-))
-EOF
 
 # Cleanup
 RUN rm -rf libs/download libs/build src/node_modules R/download
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,9 +18,10 @@ RUN mkdir -p /etc/apt/keyrings && \
     apt-get update && \
     apt-get install nodejs -y
 
-RUN curl -L https://rig.r-pkg.org/deb/rig.gpg -o /etc/apt/trusted.gpg.d/rig.gpg &&\
-    sh -c 'echo "deb http://rig.r-pkg.org/deb rig main" > /etc/apt/sources.list.d/rig.list' &&\
-    apt update && apt install r-rig
+RUN curl -L https://rig.r-pkg.org/deb/rig.gpg -o /etc/apt/trusted.gpg.d/rig.gpg && \
+    echo "deb http://rig.r-pkg.org/deb rig main" > /etc/apt/sources.list.d/rig.list && \
+    apt-get update && \
+    apt-get install r-rig -y
 
 # Because $HOME gets masked by GHA with the host $HOME
 ENV R_LIBS_USER=/opt/R/current/lib/R/site-library
@@ -42,6 +43,8 @@ RUN cd libs && make all
 RUN make
 
 # Cleanup
+RUN apt-get clean && rm -rf /var/lib/apt/lists/*
+RUN rm -rf /tmp/rig
 RUN rm -rf libs/download libs/build src/node_modules R/download
 RUN cd src && make clean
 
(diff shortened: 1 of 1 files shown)


georgestagg:
Nice! This looks like a great idea to me, thanks so much.

I'm just doing some local testing in Docker before I approve and merge.

jeroen:
OK cool! If `rig` does a little bit too much for your taste you can also grab the installer directly from:

```
https://cdn.posit.co/r/ubuntu-2204/pkgs/r-4.3.0_1_amd64.deb
```

But then you still have to manually set the p3m repositories.

georgestagg:
> If rig does a little bit too much for your taste you can also grab the installer directly from [...]

Hmmm. I'll take a look at how much `rig` adds to the image size vs. pulling the `.deb` directly and make a decision once I know for sure.

gaborcsardi:
> Edit: Image size before this PR: 2.94GB. After: 4.69GB. However, using the .deb directly doesn't seem to reduce the image size much, so I don't think it's rig adding the extra size.

The rstudio/r-builds R builds have a lot of dependencies that are not strictly required, e.g. compilers. The rig binary itself is less than 10MB AFAIR, and has no dependencies. 

You probably also want to delete the downloaded deb packages after `rig add`.

georgestagg:
> The rig binary itself is less than 10MB AFAIR, and has no dependencies.

Brill, thanks for confirming. 10MB is a bargain for the management benefit rig itself provides.

So, the cost is in using the rstudio R builds. I guess it's not *so* bad, and probably worth the extra GB (if you're already downloading 3GB in any case) to ensure we have a confirmed working native R of the right version.

jeroen:
Hmm are you sure those numbers are correct? I would be surprised if the posit build of R has many more dependencies than the debian buil. 

I have the version from this PR in my own ghcr, and the sizes are uncompressed:

```sh
# docker image ls ghcr.io/r-wasm/webr:main
REPOSITORY            TAG       IMAGE ID       CREATED      SIZE
ghcr.io/r-wasm/webr   main      3a13fe3601ba   5 days ago   2.94GB

# docker image ls ghcr.io/jeroen/webr:main
REPOSITORY            TAG       IMAGE ID       CREATED       SIZE
ghcr.io/jeroen/webr   main      8456fd927511   3 hours ago   3.17GB
```

And compressed: 

```sh
dockersize() { docker manifest inspect -v "$1" | jq -c 'if type == "array" then .[] else . end' |  jq -r '[ ( .Descriptor.platform | [ .os, .architecture, .variant, ."os.version" ] | del(..|nulls) | join("/") ), ( [ .SchemaV2Manifest.layers[].size ] | add ) ] | join(" ")' | numfmt --to iec --format '%.2f' --field 2 | column -t ; }

## jeroen@dev:~$ dockersize ghcr.io/r-wasm/webr:main
linux/amd64  1.07G

##jeroen@dev:~$ dockersize ghcr.io/jeroen/webr:main
linux/amd64  1.18G
```

Dockerfile
```diff
@@ -16,19 +16,18 @@ RUN mkdir -p /etc/apt/keyrings && \
         https://deb.nodesource.com/node_18.x nodistro main" | \
     tee /etc/apt/sources.list.d/nodesource.list && \
     apt-get update && \
-    apt-get install nodejs -y
+    apt-get install nodejs -y &&\
+    apt-get clean
```
georgestagg, line 20: We squash the docker layers right at the end of the `Dockerfile`, so IIUC I think we can get away with doing an `apt-get clean` just once during cleanup near the end.

Dockerfile
```diff
@@ -16,19 +16,18 @@ RUN mkdir -p /etc/apt/keyrings && \
...
 
-# Install standard R so that we can use PPM binaries
-RUN apt-get install -y --no-install-recommends software-properties-common && \
-    add-apt-repository --enable-source --yes "ppa:marutter/rrutter4.0" && \
-    add-apt-repository --enable-source --yes "ppa:c2d4u.team/c2d4u4.0+" && \
-    apt-get purge -y software-properties-common && \
-    apt-get autoremove -y
+RUN curl -L https://rig.r-pkg.org/deb/rig.gpg -o /etc/apt/trusted.gpg.d/rig.gpg &&\
+    sh -c 'echo "deb http://rig.r-pkg.org/deb rig main" > /etc/apt/sources.list.d/rig.list' &&\
```
georgestagg, line 23: Do we need the `sh -c` here?

```suggestion
echo "deb http://rig.r-pkg.org/deb rig main" > /etc/apt/sources.list.d/rig.list &&\
```

Dockerfile
```diff
@@ -16,19 +16,18 @@ RUN mkdir -p /etc/apt/keyrings && \
...
+    apt update && \
+    apt install r-rig && \
+    apt-get clean
```
georgestagg, line 26: IIRC we should use `apt-get` rather than `apt` in scripts.

```suggestion
    apt-get update && apt-get install r-rig -y
```

The `-y` is not strictly required here but if `r-rig` adds dependencies later it means they'll be installed too.

Dockerfile
```diff
@@ -45,21 +44,6 @@ RUN R CMD INSTALL packages/webr
...
-        getRversion(),
-        R.version["platform"],
-        R.version["arch"],
-        R.version["os"]
-    )
-))
-EOF
-
 # Cleanup
 RUN rm -rf libs/download libs/build src/node_modules R/download
```
georgestagg, line 48: Include the cleanup for `apt` and `rig` here?

```suggestion
# Cleanup
RUN apt-get clean && rm -rf /var/lib/apt/lists/*
RUN rm -rf /tmp/rig
RUN rm -rf libs/download libs/build src/node_modules R/download
```

georgestagg (approved):
Interesting! Here is how things look on my end:

Screenshot 2023-11-06 at 21 37 03

Where the untagged images are with and without rig. Perhaps the difference is because I built those locally on arm macOS, rather than through GHA/Linux? I don't really see why that should be the case, though.

In any case, I'm happy to continue with using rig and RStudio's R builds. Other than a few minor changes, let's merge and see where things lie after CI catches up.

jeroen:
Thanks. I'm reverting the last commit that added `apt-cache clean` because it did not reduce the image size at all.

jeroen:
Awesome. Going to trigger 40k rebuilds with this R-4.3.0 so that we don't get the `warning: this package was built with R-4.3.2` when loading the packages in webr.

georgestagg:
Wow, that's a lot of packages! Good luck!

gaborcsardi:
Just out of curiosity, why are you not using R 4.3.2?

jeroen:
Afaik, webr has not been ported yet to 4.3.2.

Op wo 8 nov. 2023 15:50 schreef Gábor Csárdi ***@***.***>:

georgestagg:
The current release of webR uses R 4.3.0 as the base source. I think using R 4.3.2 to build Wasm packages does work, but it shows a warning message when loading the resulting package in webR.

---

As for the reason webR still uses R 4.3.0, it's because webR patches the source code to reimplement some of R's features using web browser APIs, and also makes some other changes for compatibility with the WebAssembly and Emscripten environment. From experience, when I update the version of the R source used the patches don't apply cleanly and need to be reworked. This has been a non-trivial amount of work in the past, so I don't focus too much on following every R patch release.

Perhaps in the future, some of the patches can be upstreamed to R core to make the Wasm build process simpler.

PR #331, opened 2023-12-12
update libjpeg to fix missing .pc file

jeroen:
The `ragg` package currently cannot be built because it cannot find libjpeg. It turns out `libjpeg.pc` is missing. Updating libjpeg seems to resolve the problem.

### DIFF
--- a/libs/recipes/jpeg/rules.mk
+++ b/libs/recipes/jpeg/rules.mk
@@ -1,4 +1,4 @@
-LIBJPEG_VERSION = 9b
+LIBJPEG_VERSION = 9e
 LIBJPEG_TARBALL = $(DOWNLOAD)/libjpeg-$(LIBJPEG_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


georgestagg (approved):
Brilliant work, a one-byte change to get `ragg` working!

I still had a couple of problems getting `ragg` fully up and running, probably because I'm building on macOS rather than Linux.

I've dealt with these issues before in the form of R package forks in the `r-wasm` GH org, but we're in a better position now so I've opened a couple of PRs to upstream some fixes:

* r-lib/systemfonts#109
* r-lib/ragg#152

With the updated `libjpeg`, and these further changes above, I can confirm basic use of Ragg is running in my local development build of webR. Nice!

```r
> webr::install("ragg", repos="http://127.0.0.1:9000/repo")
Downloading webR package: cpp11
Downloading webR package: systemfonts
Downloading webR package: textshaping
Downloading webR package: ragg
> library(ragg)
> agg_png('plot.png', width = 1000, height = 1000, res = 200)
> plot(rnorm(1000), rnorm(1000))
> dev.off()
null device 
          1
```

plot (3)

PR #333, opened 2023-12-14
add libsodium

jeroen:
For the CRAN package `sodium`.


libs/recipes/sodium/rules.mk
```diff
@@ -0,0 +1,21 @@
...
+	mkdir -p $(DOWNLOAD)
+	wget $(LIBSODIUM_URL) -O $@
+
+$(LIBSODIUM_WASM_LIB): $(LIBSODIUM_TARBALL) 
+	mkdir -p $(BUILD)/libsodium-$(LIBSODIUM_VERSION)/build
+	tar -C $(BUILD) -xf $(LIBSODIUM_TARBALL)
+	cd $(BUILD)/libsodium-$(LIBSODIUM_VERSION)/build && \
+	  emconfigure ../configure \
+	    --enable-shared=no \
+	    --enable-static=yes \
```
georgestagg, line 18:
```suggestion
	    --enable-static=yes \
	    --without-pthreads \
```

With Emscripten, support for both `-pthread` and `-s SIDE_MODULE=1` at the same time is currently experimental
and does not work with R packages for webR. Without this tweak, the `sodium` package fails to load.

georgestagg (approved):
Thanks! A minor tweak is suggested below, after which I'm happy to merge.

P.S. Don't mind the CI failure in `build-nix`.

libs/recipes/sodium/rules.mk
jeroen: Good to know, thanks

PR #336, opened 2023-12-18
Add imagemagick

jeroen:
Should we disable command line utilties?


georgestagg:
As discussed, they're OK for now but we'll return and disable them if the release build times become too long.

PR #338, opened 2023-12-18
[WIP] poppler

jeroen:
We need this for the `pdftools` package.

This does not work yet, because `emcmake` seems to be unable to find freetype.


jeroen:
Sorry messed up a branch, started new PR

PR #345, opened 2023-12-24
Test build R-4.3.2

jeroen:
Minimal set of changes to test with R-4.3.2. Updated one patch and removed another one that was fixed upstream.

Note to self:
 - [x] check for regressions in `capabilities()`


georgestagg:
I tested this change on my side, and it seems to be working great. All our tests pass with my build, including the base R package tests,

```
 PASS  tests/packages/webr.test.ts (115.393 s)
  ✓ The webr R package is installed (30 ms)
  Run R default package examples and tests
    ✓ datasets (2690 ms)
    ✓ tools (12297 ms)
    ✓ utils (4547 ms)
    ✓ grDevices (31434 ms)
    ✓ graphics (4735 ms)
    ✓ stats (14863 ms)
    ✓ stats4 (992 ms)
    ✓ splines (3277 ms)
    ✓ methods (7312 ms)
    ✓ compiler (7148 ms)
    ✓ base (9164 ms)
    ✓ grid (10791 ms)
```

So, let's merge this and see how things fare on `webr.r-wasm.org/latest/` for a little while. Then, when it comes to release, we'll bump webR's own version to 0.3.0 to indicate the change.

I'll probably push one or two minor cleanup commits, then merge.

Thanks!

PR #358, opened 2024-02-11
Install Rust toolchnain

yutannihilation:
Fix https://github.com/r-wasm/rwasm/issues/20

This pull request adds two things:

1. Install the nightly Rust toolchain and `wasm32-unknown-emscripten` target
2. Install fake DEB packages of `rustc` and `cargo` in order to prevent another Rust toolchain from being installed via `apt`.

The second point is optional, so if you prefer not adding this, I'll remove.

You can inspect the docker image built from this Docker file by:

```sh
docker pull ghcr.io/yutannihilation/webr-rust:main
```

~~(I'm rebuilding the image at the time of creating this pull request as I tweaked some orders. I think the change won't matter, but just for making sure...)~~ done!

<details>

```
$ docker manifest inspect ghcr.io/yutannihilation/webr-rust:main | grep size
                "size": 4393,
                        "size": 30439111,
                        "size": 629569132,
                        "size": 718266714,
                        "size": 32,
```
= 1.38GB

</details>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,23 @@ ENV WEBR_ROOT="/opt/webr"
 ENV EM_NODE_JS="/usr/bin/node"
 ENV EMFC="/opt/flang/host/bin/flang-new"
 
+# Step 1: Build fake Rust DEB packages
+# 
+# Installing this makes sure the toolchain installed in the later step is used
+# instead of the one that the distro (Ubuntu) provides.
+FROM webr as deb_build
+RUN mkdir /opt/fake_rust/ && \
+    apt-get update && \
+	apt-get install -y equivs lsb-release &&\
+	equivs-control fake_rust && \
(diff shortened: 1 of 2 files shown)


georgestagg:
Thanks again for this!

yutannihilation:
Thank you!

By the way, do you know when this propagates to R-universe? I want to confirm Rust-powered R packages get successfully built on R-universe, but it seems the image is not built so often.

https://github.com/r-universe-org/build-wasm/pkgs/container/build-wasm/versions

georgestagg:
> By the way, do you know when this propagates to R-universe?

I don't, sorry. It will depend on when they update the Docker image in their build infrastructure.

yutannihilation:
I see, thanks. I'll ask on https://github.com/r-universe-org/help.

PR #377, opened 2024-03-06
Update plotting.qmd  - bad import in code example "Capturing Plots"

drgomulka:
The code for the example "Capturing Plots" throws the error  "Uncaught ReferenceError: WebR is not defined". 
 https://docs.r-wasm.org/webr/latest/plotting.html#capturing-plots
The line 
 import { Console } from 'https://webr.r-wasm.org/latest/webr.mjs';
should be:
 import { WebR } from 'https://webr.r-wasm.org/latest/webr.mjs';


georgestagg:
Thanks!

PR #399, opened 2024-04-02
Pass arguments to `webr::canvas()` graphics device during `captureR()`

### LINKED ISSUE #390: Setting image height/width under new capture output
With webR v0.3.0+, the graphics for `shelter.captureR()` were moved into a new `.images` slot.

However, it doesn't seem like there is a way to constrain image height and/or image width with a pre-call to `webr::canvas()` since the shelter makes its own call.

https://github.com/r-wasm/webr/blob/cc95727742fb9324bfb35fed78480ced9648be49/src/webR/webr-worker.ts#L683-L688

So, pre-calling the `webr::canvas(width=, height=)` before `shelter.captureR()` seems to have no effect. However, if I move the canvas call into the shelter evaluation, then the graphs are no longer stored inside of `.images`, but can be retrieved through reading messages after flushing, e.g. the old way from webR v0.2.x.

For a quick example, please consider the [Capturing plots](https://docs.r-wasm.org/webr/latest/plotting.html#capturing-plots) documentation entry with only a slight modification to `webr::canvas(height=400, width=320)`:

```html
<html>
  <head>
    <title>WebR Test Graph with Dimensions</title>
  </head>
  <body>
    <div id="plot-output"></div>
    <div>
      <pre><code id="out">Loading webR, please wait...</code></pre>
    </div>

    <script type="module">
      import { WebR } from 'https://webr.r-wasm.org/latest/webr.mjs';
      const webR = new WebR();
      await webR.init();

      const shelter = await new webR.Shelter();

      // <---
      // Attempt to constrain graphics output
      webR.evalRVoid('webr::canvas(width=320, height=400)');
      // ---->

      const capture = await shelter.captureR("demo(graphics)");
      capture.images.forEach((img) => {
        const canvas = document.createElement("canvas");
        // set canvas size
        canvas.width = img.width;
        canvas.height = img.height;
        
        // Draw image
        const ctx = canvas.getContext("2d");
        ctx.drawImage(img, 0, 0, img.width, img.height);
        document.getElementById("plot-output").appendChild(canvas);
      });

      shelter.purge();
    </script>
  </body>
</html>
```

Note this yields a series of `canvas` objects with a width and height of `1008 x 1008` instead of the desired `320 x 400` per image

Screenshot of the modified example test to set width and height of image generated

Thus, I think the `.images` slot contains `img` objects that do not respect the user supplied width and height values.

georgestagg:
Fixes #390.

@coatless What do you think of this implementation? Here in addition to accepting `captureGraphics: true` or `false`, optional arguments can be passed to the graphics device in the following way:

``` javascript
const shelter = await new webR.Shelter();
const capture = await shelter.captureR("hist(rnorm(1000))", {
  captureGraphics: {
    width: 504,
    height: 252,
  }
});
```

Passing an object here implies `capture: true`. This should allow for evaluation-level setting of graphics device options without having to start the device manually in the provided R code.

### DIFF
--- a/src/webR/webr-chan.ts
+++ b/src/webR/webr-chan.ts
@@ -72,8 +72,17 @@ export interface EvalROptions {
   captureConditions?: boolean;
   /**
    * Should a new canvas graphics device configured to capture plots be started?
+   * Either a boolean value, or an object with properties corresponding to
+   * `webr::canvas()` graphics device arguments.
+   * Default: `true`.
    */
-  captureGraphics?: boolean;
+  captureGraphics?: boolean | {
+    width: number;
+    height: number;
+    pointsize?: number;
+    bg?: string;
+    capture?: true;
+  };
   /**
    * Should the code automatically print output as if it were written at an R console?
    * Default: `false`.
--- a/src/webR/webr-worker.ts
+++ b/src/webR/webr-worker.ts
@@ -646,23 +646,26 @@ function captureR(expr: string | RObject, options: EvalROptions = {}): {
   output: RList,
   images: ImageBitmap[],
 } {
-  const prot = { n: 0 };
-  try {
-    const _options: Required<EvalROptions> = Object.assign(
-      {
-        env: objs.globalEnv,
-        captureStreams: true,
-        captureConditions: true,
-        captureGraphics: typeof OffscreenCanvas !== 'undefined',
-        withAutoprint: false,
-        throwJsException: true,
-        withHandlers: true,
-      },
-      replaceInObject(options, isWebRPayloadPtr, (t: WebRPayloadPtr) =>
-        RObject.wrap(t.obj.ptr)
-      )
-    );
+  const _options: Required<EvalROptions> = Object.assign(
+    {
+      env: objs.globalEnv,
+      captureStreams: true,
+      captureConditions: true,
+      captureGraphics: typeof OffscreenCanvas !== 'undefined',
+      withAutoprint: false,
+      throwJsException: true,
+      withHandlers: true,
+    },
+    replaceInObject(options, isWebRPayloadPtr, (t: WebRPayloadPtr) =>
+      RObject.wrap(t.obj.ptr)
+    )
+  );
 
+  const prot = { n: 0 };
+  const devEnvObj = new REnvironment({});
+  protectInc(devEnvObj, prot);
+
+  try {
     const envObj = new REnvironment(_options.env);
     protectInc(envObj, prot);
     if (envObj.type() !== 'environment') {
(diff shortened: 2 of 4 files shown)


src/webR/webr-worker.ts
```diff
@@ -680,11 +681,17 @@ function captureR(expr: string | RObject, options: EvalROptions = {}): {
...
+        capture: true
+      }, _options.captureGraphics));
+
       parseEvalBare(`{
         old_dev <- dev.cur()
-        webr::canvas(capture = TRUE)
+        do.call(webr::canvas, canvas_options)
         new_dev <- dev.cur()
         old_cache <- webr::canvas_cache()
+        plots <- numeric()
```
coatless, line 694: Why is `plots` being defined as an empty `numeric()`? 

The plot variable is defined later to obtain a canvas difference. 

https://github.com/r-wasm/webr/pull/399/files#diff-9a2eb982491b1c7ae12ccbc08cd97db6d219f7c4f86c25c2aebf438a0ecbba59R742

src/webR/webr-worker.ts
```diff
@@ -680,11 +681,17 @@ function captureR(expr: string | RObject, options: EvalROptions = {}): {
...
 
+      // User supplied canvas arguments, if any. Default: `capture = TRUE`
+      devEnvObj.bind('canvas_options', Object.assign({
+        capture: true
+      }, _options.captureGraphics));
+
       parseEvalBare(`{
         old_dev <- dev.cur()
-        webr::canvas(capture = TRUE)
+        do.call(webr::canvas, canvas_options)
```
coatless, line 691: This is great! Thanks for including a read of `webr::canvas()` options. 🎉

src/webR/webr-chan.ts
```diff
@@ -72,8 +72,17 @@ export interface EvalROptions {
   captureConditions?: boolean;
   /**
    * Should a new canvas graphics device configured to capture plots be started?
+   * Either a boolean value, or an object with properties corresponding to
+   * `webr::canvas()` graphics device arguments.
+   * Default: `true`.
    */
-  captureGraphics?: boolean;
+  captureGraphics?: boolean | {
```
coatless, line 79: Perfect fix _with_ backwards compatibility! 👍

coatless:
@georgestagg loving the new graphics capture options _with_ built-in backward compatibility. 

Thanks for even adding in a check to fetch`webr::canvas()`!

src/webR/webr-worker.ts
georgestagg: This is some defensive programming. If something goes wrong after this point but before where `plots` is actually defined for real, the `finally` block will still run `webr::canvas_destroy()`. With this an empty numeric vector exists for its argument, meaning we get a no-op rather than an internal error.

Issue #254, opened 2023-08-20
webr type declaration file issue


WillemSleegers:
I get the following warning in VS Code:

```
Could not find a declaration file for module 'webr'. '/Users/willem/GitHub/distributions-sampler/node_modules/webr/dist/webr.mjs' implicitly has an 'any' type.
  There are types at '/Users/willem/GitHub/distributions-sampler/node_modules/webr/dist/webR/webr-main.d.ts', but this result could not be resolved when respecting package.json "exports". The 'webr' library may need to update its package.json or typings.
```

Should the package.json file be updated?

derrickstaten:
I'd love to be able to use this in my own typescript project as well :) I've manually updated the `package.json` file to:
```
  "types": "dist/webR/webr-main.d.ts",
  "exports": {
    ".": {
      "node": "./dist/webr.cjs",
      "default": "./dist/webr.mjs",
      "types": "./dist/webR/webr-main.d.ts"
    },
```
This seems to resolve the issue. This is a temporary solution on my end as I'll have to keep manually patching the `package.json` file every time it gets reset to the distributed version.

georgestagg:
Feel free to open a PR with the changes to `package.json` for your environment. If everything continues to work on my end, I'd be happy to merge the required changes.

derrickstaten:
I'm not confident that this is the solution yet. Once I am, I will open a PR.

derrickstaten:
I've opened https://github.com/r-wasm/webr/pull/404, as it resolves warnings in a personal project

PR #404, opened 2024-04-05
change how package.json specifies types file

derrickstaten:
In a personal project, tweaking `package.json` led my IDE to start recognizing the types provided in webR. Without this I get compile-time warnings in the IDE.

I'm inexperienced, so others should double check this.

### DIFF
--- a/src/package.json
+++ b/src/package.json
@@ -38,20 +38,17 @@
     "dist"
   ],
   "main": "dist/webr.mjs",
-  "types": "dist/webr/webr-main.d.ts",
+  "types": "dist/webR/webr-main.d.ts",
   "exports": {
     ".": {
+      "types": "./dist/webR/webr-main.d.ts",
       "node": "./dist/webr.cjs",
-      "types": "./dist/webr/webr-main.d.ts",
       "default": "./dist/webr.mjs"
     },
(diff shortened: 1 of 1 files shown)


WillemSleegers:
This throws an error for me when I use it in a Next.JS project; I had to reorder the exports so the default option is the last one.

derrickstaten:
Oh interesting. Again I'm a novice here and have absolutely no clue why order of key-value pairs would matter there. 

Would you like me to update the PR with `default` as the last key inside the `exports` object?

WillemSleegers:
That makes sense to me but I should note I'm also no expert!

derrickstaten:
Done! Hopefully this works well on your end too.

georgestagg:
Thanks for this!  Fixes #254.

I've made a couple of tweaks:
 * `webr` -> `webR`
 * Move `types` property to the top of the list [1].
 * Added a similar setup for `webr/chan/serviceworker` export.
 
Hopefully this still works OK for you both. I'll merge once the GitHub Actions CI is happy, but feel free to continue the thread here if any further issues crop up in your editors.

[1] The TypeScript documentation says ["The "types" condition should always come first in "exports".](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html)

PR #445, opened 2024-06-17
Add `IDBFS` persistent storage under `PostMessage` communication channel

### LINKED ISSUE #442: Support `IDBFS` for persisting installed packages
I am working through how to persist installed R packages in a manner that can be supported outside of the Node runtime. 

The two options provided are WORKERFS and NODEFS. In my case, as I am using WebR via Rust & WASM, I cannot use the NODEFS. 

The issue with WORKERFS is that the when a page is reloaded, the workers die and the persistence with it. 

Alternatives would be to use the browser [Filesystem API](https://developer.mozilla.org/en-US/docs/Web/API/FileSystem) or IndexedDB. At present, Emscripten supports only IndexedDB via [IDBFS](https://emscripten.org/docs/api_reference/Filesystem-API.html#filesystem-api-idbfs).

Being able to support one of these would (I believe) allow me to persist R packages across page reloads.

georgestagg:
Context: #56, #442.

Closes #442.


JosiahParry:
Cruising!!!! Thank you!

Issue #431, opened 2024-06-01
webr support for {gt} package 


rainer-rq-koelle:
Congrats! I am just fascinated by what can be done with wasm/webr. I embrace tinkering with it and see a huge potential for teaching/demoing "many things R coding" related. However, I admit that the "magic of its working" is outside my comprehension. 

Following some of the tutorials/introductory material, I wanted to use webr to showcase {gt}.     
When setting up the Quarto document and rendering it, I hoped that the following would work like a charm. I have failed to find any meaningful pointer(s). 

Question on stackoverflow: https://stackoverflow.com/questions/78534930/webr-and-gt-throws-error-error-browser-must-be-a-non-empty-character-stri

The error message thrown from the web-r chunk output reads:  ```Error: 'browser' must be a non-empty character string```

Quarto yaml:
---
engine: knitr
filters:
  - webr
webr:
  packages: ['dplyr','gt']
---

And then let's trial {gt}

```{webr-r}
my_tbl <- mtcars[1:10,] |> tibble()

my_tbl |> gt()
```
<img width="620" alt="Screenshot 2024-06-01 at 09 21 34" src="https://github.com/r-wasm/webr/assets/17096612/5df91905-8d14-45f1-af53-fcc9f74ef81f">

Calling `?gt` in the webr-container on a rendered webpage results in the standard (embedded) help page for {gt}. Thus, the package {gt} appears to be loaded and available in the {webr} session.

Could someone put me on the right track? Thanks.

timelyportfolio:
@rainer-rq-koelle I am not on the official team, but I do have some experience with `html` and `R`, so I'll try to chime in with some hopefully helpful context.  This issue relates to the fact that `webr` does not have a `browseURL` equivalent viewer (such as RStudio Viewer tab in RStudio or chrome/edge/firefox).

The [`gt:::print.gt_tbl`](https://github.com/rstudio/gt/blob/master/R/print.R#L36-L46) method in an interactive session will effectively create an `html` file which then opens with `browseURL` except we do not have a browser in `webr`.  For instance, `htmlwidgets` use a very similar mechanism, and these also will not work in a webr Quarto context.

[...]

Issue #384, opened 2024-03-13
Feature Request: A viewer in the REPL


gitdemont:
Hello,

Would it be possible to have a viewer in the REPL (e.g. an iframe) listening to current in-browser session ?
So as for instance to run shiny apps alike https://github.com/georgestagg/shiny-standalone-webr-demo, display DT::datatable, ... html.

gitdemont:
Hi @georgestagg ,

I just tried your #449 PR. And it's :rocket: !

I tested a basic shiny::runGadget but without success (I was too much in a hurry)
console says it's listening to url:port but access to url:port is denied
in addition, REPL is blocked and webr.r-wasm.org/latest needs to be refreshed 

However, I was able to display several widgets (from `visNetwork` and `DT`) directly in the REPL 👍 .

Thanks a lot !

georgestagg:
Closing, with #449 webR should now cover most requirements for viewing HTMLwidgets and UI elements. Shiny and `shiny::runGadget` is a harder problem, due to the requirement of communication with the shiny server.

I think I would consider displaying Shiny apps and gadgets out of scope for the webR application. For that use case, one should use https://shinylive.io/r/, which already has the infrastructure in place to handle the Shiny server under Wasm.

gitdemont:
Thanks for the addition !

Issue #295, opened 2023-10-13
Thoughts on shimming browseURL()


ColinFay:
The browseURL() function defines : 

```
url: a non-empty character string giving the URL to be loaded. Some platforms also accept file paths.
```

As far as I can tell, this can be shimmed via running JS code in the main thread.

```
# in webr
webr::browseURL(url)
```

Two case: 

- we want to browse an external URL 

```
windows.open(url);
```

- we want to browse a local HTML file 

```
# write file from the R instance 
await webR.evalR('write("<h1>Hello world</h1>", "hello.html")');
const html = await webR.FS.readFile("/home/web_user/hello.html");
const html_content = new TextDecoder().decode(html);
var newWindow = window.open();
newWindow.document.write(html_content);
```

Possibly, we might start with supporting only remote URL (the helps says "Some platforms also accept file paths.", so I feel like this is ok to not support local file in a first time). 

I'm not exactly sure how to send JS code from webr to be executed in the main thread, though, any help on how to do this would be awesome :) 

Thanks

georgestagg:
Yes, we will definitely want to handle `utils::browseURL()` at some point. I am unsure as of yet whether it's better to handle this through patching R's source or shimming using the `webr` package. If the implementation becomes complex, with several moving parts, it might be more elegant to patch R itself directly.

One issue I can think of is that if the local HTML file depends on other files on the VFS (e.g. scripts or images), those too would need to be transferred to the main thread for display. It would probably be possible to parse the HTML file in question to look for dependencies, but that might not be the best solution.

Making use of a JS Service Worker might be possible and prefereable. This is how Shinylive exposes a web server in the R process to the main thread. However, only one service worker can be active at a time for a particular origin, so such a method will need to be mindful of the existing [webR Service Worker communication channel](https://github.com/r-wasm/webr/blob/main/src/webR/chan/serviceworker.ts) that may or may not already be running in the page.

---

> I'm not exactly sure how to send JS code from webr to be executed in the main thread

The easiest way is probably through a custom webR channel message. Running the following in R:

```r
webr::eval_js("chan.write({ type: 'myEvent', data: { mydata: 123 } })")
```

will cause a custom message to appear in [`webR.flush()`/`webR.read()`](https://docs.r-wasm.org/webr/latest/communication.html#messages) on the main thread. This can be used to pass messages and information from the R process to be handled in the main thread JavaScript thread. This mechanism is how documentation display is handled, for example, in:
[...]

PR #449, opened 2024-06-24
A HTML widget viewer for the webR application, similar to `browseURL()`

georgestagg:
A function `webr::viewer_install()` is added to the webR support package. The function sets R up so as to generate an output message over the webR communication channel when a URL viewer is invoked using `options("viewer")`.

Printing a HTML element or HTML widget in the webR application app now shows the resulting HTML output in an embedded viewer `iframe`, meaning packages like `gt()` can now work.

c.f. #295, #384, #431.

----

Screenshot 2024-06-24 at 09 30 48

----

Over the months I have considered several ways to handle requests to `browseURL()` from webR. After pondering for enough, in the interests of pragmatism this PR contains a "first-pass" approach that replaces relative JS and CSS sources in the output HTML file with content directly (encoded in base64) then displays the result in an `iframe`. With thanks to @timelyportfolio, as this is an improvement to their basic implementation method [outlined here](https://www.jsinr.me/2024/01/10/selfcontained-htmlwidgets/).

In the future, one or more of the following should be done, in order of "correctness":

* Other types of relative URL in the output should be found and replaced.
* Proper XML parsing should be done on the output, not regex replacement.
* The output should instead be made self-contained in a more robost way, e.g. using Pandoc compiled for wasm.
* The R object should instead be send to the main thread, not the HTML output. That way adding HTML dependencies to the page can be handled better in JavaScript.
* A JS Service Worker should be started, serving the `/tmp` directory containing the HTML and support files. With this, relative URLs would "just work" without any modification.


timelyportfolio:
@georgestagg Wow, this is absolutely amazing.  I'll dive deeply over the week keeping in mind the list of "to-do".  Let me know if there is anything I can do to help.

PR #467, opened 2024-08-05
Handle `WebAssembly.RuntimeError` by closing communication channel

### LINKED ISSUE #464: WebR OOM handing
A real world example of badly behaved code that will make webR crash with a stack overflow and oom:

```r
install.packages('LTMRdata', repos = c('https://sbashevkin.r-universe.dev/', 'https://repo.r-wasm.org/'))
install.packages('jsonlite')
data(Salvage, package = 'LTMRdata')
json <- jsonlite::toJSON(Salvage)
```

In node-20 this errors with

```
table index is out of bounds
```

But it has also corrupted all webR instances in the process until we call `webr.close()` on the borked session. Trying to run `evalRVoid("1+1")` on this or any other webr instance now errors with:

```
memory access out of bounds
```

Ideally when this  webR would catch the error `memory access out of bounds` and automatically `close()` itself to free up the stack from the JS process. If there is any option to set the stack limit for a single webr instance, that might help as well.

georgestagg:
This PR makes two changes:

1) If `WebAssembly.RuntimeError` is raised and bubbles all the way up to the top level, catch it and report the problem as unrecoverable on `console.error()`. Then close the communication channel, terminating the worker thread.

2) If a communication channel has been closed, reject any promises trying to write on the channel with a `WebRChannelError`.

Closes #464, because the OOM error is a `WebAssembly.RuntimeError`.

### DIFF
--- a/src/tests/webR/webr-main.test.ts
+++ b/src/tests/webR/webr-main.test.ts
@@ -1034,6 +1034,11 @@ test('Close webR communication channel', async () => {
   // Close the channel
   tempR.close();
   await expect(closedPromise).resolves.toEqual(true);
+
+  // Writing messages after closing the channel is an error
+  expect(() => tempR.writeConsole('foo <- 123')).toThrow(
+    "The webR communication channel has been closed"
+  );
 });
 
 test('Default and user provided REnv properties are merged', async () => {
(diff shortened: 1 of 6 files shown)


jeroen:
Thanks! Going to test this right away.

georgestagg:
Hopefully the failure behaviour is more reasonable now!

BTW you might need to wait a little while for CI to catch up, it rebuilds absolutely everything from scratch and takes a while before uploading the new Wasm binaries to AWS.

PR #472, opened 2024-08-15
add dark mode option to Quarto site

luisDVA:
added dark theme to yml to enable toggle between cosmo and darkly
fonts will be inconsistent but @georgestagg is OK with this (dev day pers. comm.)


georgestagg:
Thanks!

PR #473, opened 2024-08-15
Add docs note about mobile browser memory limits

### LINKED ISSUE #346: Note about limited memory allocation by wasm for webR in browsers on mobile devices
When I first started using webR when it came out, I had noticed that I couldn't analyze as large datasets while using webR on my mobile device (iPad) as compared to when I accessed webR on my desktop (Mac). I didn't think much about this for over a year until I recently started working on a mobile project that queries an API and uses webR as a dependency. I was attempting to download an .xml file and needed an xml parser; initially, I tried using the xmlconvert package but kept getting an out of memory error in R when trying to convert xml_to_df (yet worked find trying to convert the same .xml file to df on webR running on my Mac; I was able to instead use the xml2 package which I presume uses less memory). I thus did some further research and learned from this Unity forum post that some browsers have not kept up with wasm and have not allowed wasm to allocate more than 500MB of memory for wasm projects running in their browsers:

https://forum.unity.com/threads/android-chromium-unable-to-grow-allocated-memory-above-256mb-confirmed.1014475/

That forum discussion also included a test link to estimate how much memory can be allocated to wasm projects in specific browser/device combinations:

http://clb.confined.space/dump/mem_growth.html

Interestingly, when using the above test link on my 8GB iPad Pro, my iPad browsers only allocate 250-500MB max. My Mac allocated 1GB. 

When my colleague tried the above test site on his Android devices, he was only able to allocate 500MB using Google but could allocate 2GB using Firefox on his Android phone.(Unfortunately, Firefox on Apple mobile devices doesn't actually use Firefox's Gecko engine [Webkit instead]) 

Wanted to alert the webR community of this current memory limitation when using webR with certain browsers on certain mobile devices. (Fyi)

thomascwells:
Fixes #346 by adding note to documentation main page about memory limits on mobile device browsers.

### DIFF
--- a/src/docs/index.qmd
+++ b/src/docs/index.qmd
@@ -15,6 +15,10 @@ WebR makes it possible to run R code in the browser without the need for an R se
 The webR project is under active development, and the API is subject to change. Be aware that if you are using the latest build of webR the contents of this documentation might be out of date.
 :::
 
+::: callout-note
+Some browsers (especially mobile browsers) may place hard caps on the amount of RAM accessible by WebAssmebly *regardless of the amount of memory available on the host device*. Please keep this in mind when developing webR intended for use on mobile devices. Consider including a warning to mobile users if you expect your webR application to exceed 500 MB.
+:::
+
 ## Try it out
 
 ### WebR REPL
--- a/src/docs/index.qmd
+++ b/src/docs/index.qmd
@@ -16,7 +16,7 @@ The webR project is under active development, and the API is subject to change. 
 :::
 
 ::: callout-note
-Some browsers (especially mobile browsers) may place hard caps on the amount of RAM accessible by WebAssmebly *regardless of the amount of memory available on the host device*. Please keep this in mind when developing webR intended for use on mobile devices. Consider including a warning to mobile users if you expect your webR application to exceed 500 MB.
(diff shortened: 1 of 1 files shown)


thomascwells:
It might be better to have a section of the documentation discussing general limitations and considerations for webR and WASM or possibly a section about mobile-devices specifically. If it would be better to have a section like that, I'm happy to move this there, but this seemed like the easiest quick fix.

src/docs/index.qmd
```diff
@@ -15,6 +15,10 @@ WebR makes it possible to run R code in the browser without the need for an R se
 The webR project is under active development, and the API is subject to change. Be aware that if you are using the latest build of webR the contents of this documentation might be out of date.
 :::
 
+::: callout-note
+Some browsers (especially mobile browsers) may place hard caps on the amount of RAM accessible by WebAssmebly *regardless of the amount of memory available on the host device*. Please keep this in mind when developing webR intended for use on mobile devices. Consider including a warning to mobile users if you expect your webR application to exceed 500 MB.
```
georgestagg, line 19:
```suggestion
Some browsers (especially mobile browsers) may place restrictive limits on the amount of RAM provided to WebAssmebly *regardless of the amount of memory available on the host device*. Please keep this in mind when developing applications with webR intended for use on mobile devices.
```

* Minor rewording.

* I don't think we need to include the last sentence. We shouldn't include any specific memory limits here since they will likely differ between mobile devices and can change over time as browsers are developed.

georgestagg (approved):
Thanks and agreed, this is a good start. Just a small minor suggestion below and we can merge!

PR #477, opened 2024-09-05
Add support for mounting `.tgz` files with filesystem metadata appended

georgestagg:
The new format is made the default when installing binary Wasm R packages.

Also fixes mounting filesystem images under Node, and so additional unit tests have been added for mounting both `.data`/`.js.metadata` formatted packages from Emscripten's `file_packager` tool, and our own metadata appended `.tar` format.

Tagging @jeroen, for info.


src/webR/mount.ts
```diff
@@ -0,0 +1,168 @@
...
+  const len = view.getUint32(view.byteLength - 4);
+
+  if (magic !== 2003133010 || block === 0 || len === 0) {
+    throw new Error("Can't mount archive, no VFS metadata found.");
+  }
+
+  const bytes = new DataView(buffer, 512 * block, len);
+  const metadata = JSON.parse(new TextDecoder().decode(bytes)) as FSMetaData;
+  return { data: buffer, metadata };
+}
```
georgestagg, line 168: This is the primary function that extracts the appended filesystem metadata from the end of the `.tar` data.

PR #481, opened 2024-09-13
Fix a typo in the docs

christianp:


georgestagg:
Thanks!

christianp:
This was simple, but you can take its presence as an omen of a much gnarlier issue that I'm about to file!

(unless I spot a bug in my code)

PR #485, opened 2024-09-20
Add workaround for old freetype

### LINKED ISSUE #480: Patch conflicting TYPEOF macro in freetype
The version of freetype that is vendored with emscripten defines a `TYPEPOF` macro which conflicts with the R macro. We [hacked around this](https://github.com/r-lib/systemfonts/pull/109) but it has [surfaced again](https://github.com/r-universe/r-lib/actions/runs/10805477703/job/29972749632) in the devel version of cpp11.

New versions of freetype have renamed this macro, but there is probably a reason that emscripten has not upgraded.

So we can try to patch it in our toolchain, or ship our own build of freetype.

jeroen:
Workaround for the TYPEOF macro conflict in old freetype versions. This will enable to build the `systemfonts` R package again, until we have upgraded the emscripten stack.

Fixes #480

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -94,6 +94,9 @@ RUN rm -rf /tmp/rig
 RUN rm -rf libs/download libs/build src/node_modules R/download
 RUN cd src && make clean
 
+# Temp workaround for https://github.com/emscripten-core/emscripten/issues/22571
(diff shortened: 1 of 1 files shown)


georgestagg:
LGTM, thanks!

PR #487, opened 2024-09-23
Add HDF5 1.12.3

dipterix:


dipterix:
Hi I would like to get `hdf5r` compiled on webr repo. Is there any way I can test if this PR has all the components needed to install that package?

* The HDF5 version is 1.12.3 because this is a version that both `hdf5r` and `rhdf5` support. 
* In the [hdf5/targets.mk](https://github.com/dipterix/webr/blob/114da292adb095c76c5ae2000ade91a709f6c7f6/libs/recipes/hdf5/targets.mk#L3), I don't know values to assign to `OPTIONAL_HDF5_BINS`. I put `h5cc` because it's the executable program filename.

libs/recipes/hdf5/targets.mk
```diff
@@ -0,0 +1,3 @@
+HDF5_WASM_LIB = $(WASM)/lib/libhdf5.a
+OPTIONAL_HDF5_LIBS += $(HDF5_WASM_LIB)
+OPTIONAL_HDF5_BINS += h5cc
```
georgestagg, line 3:
```suggestion
OPTIONAL_WASM_LIBS += $(HDF5_WASM_LIB)
OPTIONAL_WASM_BINS += h5cc
```

These targets should be named as above, so that they're automatically included when running `make all`.

libs/recipes/hdf5/rules.mk
```diff
@@ -0,0 +1,28 @@
...
+	mkdir -p $(BUILD)/hdf5-$(HDF5_VERSION)/build
+	tar -C $(BUILD) -xf $(HDF5_TARBALL)
+	cd $(BUILD)/hdf5-$(HDF5_VERSION)/build && \
+	emconfigure ../configure \
+		--enable-build-mode=production \ 
+		--disable-dependency-tracking \ 
+		--disable-silent-rules \ 
+		--enable-shared=no \
+		--enable-static=yes \
+		--prefix=$(WASM) && \
```
georgestagg, line 26:
```suggestion
	emconfigure ../configure \
	  --enable-build-mode=production \
	  --disable-dependency-tracking \
	  --disable-silent-rules \
	  --enable-shared=no \
	  --enable-static=yes \
	  --prefix=$(WASM) && \
```

Some of these lines have trailing spaces after the `\`, breaking the line continuation. I've also tweaked the indendation slightly.

libs/recipes/hdf5/rules.mk
```diff
@@ -0,0 +1,28 @@
...
+$(HDF5_TARBALL):
+	mkdir -p $(DOWNLOAD)
+	wget $(HDF5_URL) -O $@
+
+# How to include --with-szlib=zlib?
+
+$(HDF5_WASM_LIB): $(HDF5_TARBALL)
+	mkdir -p $(BUILD)/hdf5-$(HDF5_VERSION)/build
+	tar -C $(BUILD) -xf $(HDF5_TARBALL)
+	cd $(BUILD)/hdf5-$(HDF5_VERSION)/build && \
```
georgestagg, line 19:
```suggestion
	mkdir -p $(BUILD)/hdf5-$(HDF5_VERSION)/hdfsrc/build
	tar -C $(BUILD)/hdf5-$(HDF5_VERSION) -xf $(HDF5_TARBALL)
	cd $(BUILD)/hdf5-$(HDF5_VERSION)/hdfsrc/build && \
```

This handles the extra `./hdfsrc` directory in source tarball.

georgestagg (changes requested):
Hi, thanks for taking the time to write this PR!

It looks like the hdf5 sources are not being built yet, because the make variables need to be named `OPTIONAL_WASM_LIBS` and `OPTIONAL_WASM_BINS` to be included in the `make all` command. Unfortunately, that means the tests have passed but not actually tested your changes.

Once I made that change, some issues have arisen. You should be able to see this yourself by renaming the make variables as suggested, then running:

```
cd libs; make all
```

I have made a first pass at correcting the issues, and included the required changes to get building up and running in this review.

I find that the build currently fails with the following:

```
LD_LIBRARY_PATH="$LD_LIBRARY_PATH`echo -L/Users/georgestagg/work/webr/wasm/lib -s USE_BZIP2=1 -s USE_ZLIB=1 -Oz -fwasm-exceptions -s SUPPORT_LONGJMP=wasm |                  \
                sed -e 's/-L/:/g' -e 's/ //g'`"                               \
         ./H5make_libsettings  H5lib_settings.c  ||                               \
            (test $HDF5_Make_Ignore && echo "*** Error ignored") ||          \
            (rm -f H5lib_settings.c ; exit 1)
/bin/sh: ./H5make_libsettings: Permission denied
make[2]: *** [H5lib_settings.c] Error 1
make[1]: *** [install-recursive] Error 1
emmake: error: 'make install' failed (returned 2)
make: *** [/Users/georgestagg/work/webr/wasm/lib/libhdf5.a] Error 1
```

Once the changes below have been applied, I will come back later and take another pass at getting hdf5 to compile for Wasm. Possibly `H5make_libsettings` is an intermediate binary and we just need to patch `configure` to run it through `node`, or something similar.

Once the resulting library is compiling and available, it should be that the R package "just works", but we will see.

georgestagg:
Is there any way to use a newer version of HDF5? I have tested and am able to build v1.14.5, because there have been come changes in the HDF5 source that much improve the process of cross-compilation.

dipterix:
Please let me check if package `hdf5r` and `rhdf5` work with the new versions. The latest header version for hdf5r is 1.12.0 and for rhdf5 is 1.10.0. I'm not sure whether newer versions will work.

Also when compiling the package, I encountered this error:

```
wasm-ld: error: unable to find library -lsz
```

Maybe I should make a PR to add `libaec` as well...

georgestagg:
Whoops, I messed that up!

Let's try again at https://github.com/r-wasm/webr/pull/540.

PR #545, opened 2025-06-08
Do not call strip in Makevars

jeroen:
Some packages use `$(STRIP_STATIC_LIB)` or `$(STRIP_SHARED_LIB)` on vendored libraries. However calling native `/usr/bin/strip` on a WebAssembly binary gives an error. So we make this a no-op.


georgestagg:
LGTM, thanks!

PR #557, opened 2025-07-08
Update Fontconfig to v2.15.0

### LINKED ISSUE #544: Update fontconfig
We have fontconfig `2.12.5` which is quite old and causing some problems for e.g. `systemfonts` https://github.com/r-universe/r-lib/actions/runs/15410167134

By comparison, RHEL 8 has fontconfig 2.13.1 and RHEL 9 has fontconfig 2.14.0.

georgestagg:
Closes #544.


jeroen:
Maybe try upgrading to fontconfig 2.14.1 or 2.15.0 while we are at it?

georgestagg:
Yeah, let's bump to 2.15.0. There's also a 2.16.0 but that looks pretty new, I don't even think the active development version of Ubuntu has that yet.

PR #561, opened 2025-07-14
Add libcurl stack

jeroen:
This disables most functionality, but at least we can build the `curl` R package and use offline functionality.

Inspired by https://github.com/ading2210/libcurl.js/blob/main/client/tools/curl.sh

Maybe update to openssl 3.5.0 and disable building docs to speed things up.


georgestagg:
Thanks!

PR #562, opened 2025-07-16
Proxy WebSocket communication through the main thread

georgestagg:
Also:
 * Adds support for `webr::eval_js(..., await = TRUE)`, based on a similar mechanism.
 * Downloads SSL cacert bundle from curl.se at build time, and installs it to the Emscripten VFS.
 * Set default WebSocket-to-TCP protocol to `wss://`.
 * Document networking with webR.

This PR also removes the service worker communication channel. This channel has been deprecated for a while now, is never chosen in `Automatic` mode, and when explicitly selected emits a warning message saying it will be deleted soon.

After this change the channel will indeed be gone and can no longer be selected at all.

---

Once everything has been setup, and a WebSocket proxy + SOCKS tunnel is running outside the browser:

```r
install.packages(c("curl", "httr2", "jsonlite"))
library(httr2)

response <- request("https://hb.cran.dev/get") |>
  req_perform() |>
  resp_body_json()

response$headers$`User-Agent`
#
# [1] "httr2/1.2.0 r-curl/6.4.0 libcurl/8.14.1"
```


georgestagg:
> Has https://webr.r-wasm.org/proxy-websockets/ been updated yet to this latest PR? 

Not yet, I trigger PR deployments manually.

I'll add the ca-bundle path config to curl and deploy a new build once I'm back next week :)

jeroen:
Here is an example pushing this to download 100 files in parallel over HTTP2:

```r
webr::eval_js("SOCKFS.websocketArgs.url = 'wss://'")

install.packages('curl')
Sys.setenv(ALL_PROXY='socks5h://ws.opencpu.org:443')
mirror <- 'https://cran.rstudio.com'
df <- read.dcf(curl::curl(sprintf('%s/src/contrib/PACKAGES', mirror)))
pkgs <- df[1:200,'Package']
urls <- sprintf('%s/web/packages/%s/DESCRIPTION', mirror, pkgs)
outdir <- file.path(tempdir(), 'descriptions')
files <- sprintf('%s/%s.txt', outdir, pkgs)
dir.create(outdir, showWarnings = FALSE)
results <- curl::multi_download(urls, files, progress = FALSE, multiplex = T, verbose=T)
all(results$status == 200)
list.files(outdir)
```

jeroen:
Thanks, all looks good now from my end. I think this is ready for public preview.

src/docs/networking.qmd
```diff
@@ -0,0 +1,135 @@
...
+---
+
+## Networking under WebAssembly
+
+WebAssembly runs in a sandbox environment and doesn't support direct network socket connections for security reasons. This means that traditional R networking and file downloading approaches need some extra work to be used with webR.
+
+Luckily, webR provides two solutions to enable network connectivity:
+
+1. **Basic downloads** using base R's `download.file()`.
+2. **Advanced networking** with a WebSocket proxy, for packages like curl and httr2.
```
jeroen, line 14:
```suggestion
2. **Advanced networking** using a WebSocket proxy, for packages like curl and httr2.
```

src/docs/networking.qmd
```diff
@@ -0,0 +1,135 @@
...
+
+For simple file downloads, webR patches the `download.file()` function to work inside a web browser. The patch works by intercepting requests and handling them using JavaScript APIs. This method is easy to use and in many cases is sufficient.
+
+### Limitations
+
+The approach works well for basic HTTP(S) downloads, but it does come with some limitations imposed by the security of the browser environment:
+
+- The server hosting the file must supply CORS ([Cross-Origin Resource Sharing](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS)) headers to permit cross-origin requests.
+
+- Only simple HTTP/HTTPS `GET` requests are supported. You cannot use `POST` requests or set certain HTTP headers.
```
jeroen, line 26:
```suggestion
- Only simple HTTP/HTTPS `GET` requests are supported. You cannot use another HTTP method, set headers, or use any of the advanced options available in curl/httr2.
```

PR #564, opened 2025-07-23
Update OpenSSL

jeroen:
OpenSSL 3.5 is the current LTS branch: https://openssl-library.org/source/

We also disable building documentation and dynamic modules.


georgestagg:
Thanks!

PR #570, opened 2025-09-19
Add shim for requireNamespace

### LINKED ISSUE #474: Shim `loadNamespace()`, `requireNamespace()` (and others?) in the same way as `library()` and `require()`
I would like `requireNamespace("dplyr")` call to download webR package. Currently that's not the case and I need to use `require("dplyr")` instead

AABoyles:
Fixes #474


georgestagg (approved):
LGTM, thanks!

PR #571, opened 2025-09-19
Add inline documentation for evalR methods

### LINKED ISSUE #436: Add inline documentation to evalR
I tried using `evalRString()` in the hopes that it would return the results as a string. However, it _expects_ a string and will error if a string is not returned from the function. 

It would be really nice if this expectation was documented in the API reference 
https://docs.r-wasm.org/webr/latest/api/js/classes/WebR.WebR.html#evalrstring

https://github.com/r-wasm/webr/blob/c07d8997ddbb58f63002e8d46792be54edcd134c/src/webR/webr-main.ts#L416-L418

AABoyles:
Contributes to but does NOT fix #436: the API documentation will need to be rebuilt and deployed.

### DIFF
--- a/src/webR/webr-main.ts
+++ b/src/webR/webr-main.ts
@@ -470,7 +470,7 @@ export class WebR {
    * Evaluate the given R code, returning a promise for no return data.
    * @param {string} code The R code to evaluate.
    * @param {EvalROptions} [options] Options for the execution environment.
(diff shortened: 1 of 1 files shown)


georgestagg (commented):
Thanks, looks great! Some minor suggestions are below.

PR #579, opened 2025-09-29
added blosc recipe

pepijn-devries:
With this addition, webr is building without problems. I'm not 100% the added recipe is correct. Please let me know if there is anything I can do to facilitate


georgestagg:
Sorry for the slow reply. I tried this locally but get,

```
emconfigure: error: '../configure --enable-shared=no --enable-static=yes --with-zstd --with-zlib --with-lz4 --prefix=/Users/georgestagg/work/webr/wasm' failed: [Errno 2] No such file or directory: '../configure'
```

Possibly the blosc project builds using `cmake` or some other alternative build system.

In #593 I have added extra workflows to test this better in CI. Once that has merged, basing your branch on `main` should run the extra checks when adding new library recipes like this.

pepijn-devries:
Thanks, I will try to fix this in my fork, and test with the new CI.

Issue #585, opened 2025-12-16
File download.file() does not follow 302 redirects in Node.js [PATCH]


HenrikBengtsson:
It looks like `download.file()` does not follow redirected URLs.

# Example

```sh
$ node webR-download.file.bug.js 
Node.js v v20.19.5
webR v0.5.7


download.file(<redirecting URL>):
Error fetching https://cran.r-universe.dev/bin/emscripten/contrib/4.5/listenv_0.10.0.tgz - Found. Redirecting to https://cdn.r-universe.dev/d9f4c99b3ea4a96ae2b5782941df7524a288a7bb26efb4c95caaca25e09bf6ac
trying URL 'https://cran.r-universe.dev/bin/emscripten/contrib/4.5/listenv_0.10.0.tgz'
Warning in download.file("https://cran.r-universe.dev/bin/emscripten/contrib/4.5/listenv_0.10.0.tgz",  :
  URL https://cran.r-universe.dev/bin/emscripten/contrib/4.5/listenv_0.10.0.tgz: Download failed. See the Javascript console for further information
Error: download from 'https://cran.r-universe.dev/bin/emscripten/contrib/4.5/listenv_0.10.0.tgz' failed


download.file(<directing URL>):
trying URL 'https://cdn.r-universe.dev/d9f4c99b3ea4a96ae2b5782941df7524a288a7bb26efb4c95caaca25e09bf6ac'
Success: TRUE
```

where `webR-download.file.bug.js`:

```js
import { WebR } from "webr";
import WebSocket from "ws";

async function webr_eval_code(code) {
    let shelter = await new webR.Shelter()
    let result = await shelter.captureR(code, {
        withAutoprint: true,
        captureStreams: true,
        captureConditions: false
    })
    for (const { type, data } of result.output) {
        if (type === "stdout") {
            console.log(data);
        } else if (type === "stderr") {
            console.error(data);
        }
    }
    shelter.purge();
    return result;
}

console.log('Node.js v', process.version);

// Launch webR instance
const webR = new WebR();
await webR.init();
console.log("webR v" + webR.version);

console.log("\n\ndownload.file(<redirecting URL>):");
await webr_eval_code([
    'download.file("https://cran.r-universe.dev/bin/emscripten/contrib/4.5/listenv_0.10.0.tgz", "foo.tgz", mode = "wb")',
    'message("Success: ", file.exists("foo.tgz"))'
]);

console.log("\n\ndownload.file(<directing URL>):");
[...]

PR #586, opened 2025-12-16
downloadFileContent(): Follow URL redirects in Node [#585]

HenrikBengtsson:


src/webR/webr-worker.ts
```diff
@@ -610,15 +614,38 @@ function downloadFileContent(URL: string, headers: Array<string> = []): XHRRespo
...
+
+    // Follow 3xx redirects
+    if (status >= 300 && status < 400) {
+      const location = IN_NODE
+        ? responseHeaders?.location
+        : request.getResponseHeader('Location');
+
+      if (location) {
+        // Resolve relative URLs against the original URL
+        const redirectUrl = new URL(location, url).href;
```
dirkschumacher: `new URL` can throw a `TypeError` exception in case the url is not valid. That should be caught here IMHO.

src/webR/webr-worker.ts
```diff
@@ -592,9 +592,13 @@ function copyFSNode(obj: FSNode): FSNode {
   return retObj;
 }
 
-function downloadFileContent(URL: string, headers: Array<string> = []): XHRResponse {
+function downloadFileContent(url: string, headers: Array<string> = [], maxRedirects: number = 10): XHRResponse {
+  if (maxRedirects == 0) {
```
dirkschumacher: Maybe change to `maxRedirects <= 0` - just in case to prevent future potential bugs. Right now it is correct though

src/webR/webr-worker.ts
dirkschumacher: The URL can be "be relative to the request URL or an absolute URL" https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Location#url

The case of an absolute URL probably creates an exception here.

src/webR/webr-worker.ts
HenrikBengtsson: Thanks. Added.

src/webR/webr-worker.ts
HenrikBengtsson: Great point. I've updated to handler any errors - including the special case TypeError. Can you please take a look?

src/webR/webr-worker.ts
```diff
@@ -610,15 +614,50 @@ function downloadFileContent(URL: string, headers: Array<string> = []): XHRRespo
...
+      const parsed = JSON.parse(String(request.status)) as {
+        data: { statusCode: number; headers: Record<string, string> }
+      };
+      status = parsed.data.statusCode;
+      responseHeaders = parsed.data.headers;
+    } else {
+      status = request.status;
+    }
+
+    // Follow 3xx redirects
```
HenrikBengtsson: The browser already takes care of redirects. Should the code for handling 3xx redirects be conditioned by `IN_NODE`, e.g.

```js
    if (IN_NODE && status >= 300 && status < 400) {
       ...
    }
 ```

?

File: src/webR/webr-worker.ts (hunk 1 above)
Comment (HenrikBengtsson) [reply src/webR/webr-worker.ts]: > The case of an absolute URL probably creates an exception here.

The example of #585 uses an absolute URL redirect.

According to https://developer.mozilla.org/en-US/docs/Web/API/URL/URL, it sounds like `url` is ignored if `location` is an absolute URL;

> A string or any other object with a stringifier that represents an absolute URL or a relative reference to a base URL. If `url` is a relative reference, `base` is required, and is used to resolve the final URL. If `url` is an absolute URL, a given `base` will not be used to create the resulting URL.

Was that your concern here?

PS. I'm not a regular JavaScript developer.

File: src/webR/webr-worker.ts (hunk 3 above)
Comment (HenrikBengtsson) [reply src/webR/webr-worker.ts]: If this is done, the code can be cleaned up even further.

File: src/webR/webr-worker.ts (hunk 1 above)
Comment (dirkschumacher) [reply src/webR/webr-worker.ts]: Correct, in that case it's seems to be fine

**Comment by dirkschumacher** [comment]
LGTM, but I am not a maintainer 😅. There are some [tests](https://github.com/r-wasm/webr/blob/main/src/tests/webR/webr-worker.test.ts) for webr-workers, so maybe there is a way to test this feature.

File: src/webR/webr-worker.ts (hunk 3 above)
Comment (HenrikBengtsson) [reply src/webR/webr-worker.ts]: I've consolidated all `IN_NODE` code to one `if (IN_NODE) { ... }` statement.

**Comment by HenrikBengtsson** [comment]
The GA build-webr checks produced:

```
/__w/webr/webr/src/webR/webr-worker.ts
Error:   595:72  error  Type number trivially inferred from a number literal, remove type annotation  @typescript-eslint/no-inferrable-types
```

I address that in commit 1f2011e. @georgestagg , could please trigger a rerun?

**Comment by georgestagg** [comment]
Yeah, the linter is configured to be quite picky…

You may be able to check it locally quicker by cd-ing into `src` and running `make lint`

**Comment by HenrikBengtsson** [comment]
> You may be able to check it locally quicker by cd-ing into `src` and running `make lint`

Thx - worked like a charm. Confirming I could reproduce the lint error, which is gone in the most recent commit.

**Review by georgestagg — APPROVED** [review]
LGTM!

If we wanted, we could add a test for `download.file()` in the same place as we already have a couple of tests that depend on networking [here](https://github.com/r-wasm/webr/blob/71fa42f80dbb7e72602df180e9500c269ab70134/src/tests/webR/webr-worker.test.ts#L40).

Happy to merge without, but it's an option if you'd like to add it. We'd need to test downloading a file from with a reasonable URL that will always be consistent and available.

**Comment by HenrikBengtsson** [comment]
I've added direct tests of `utils::download.file()` with and without URL redirects. 

I also added a test of `<shim>::install.packages()` from R-universe, which (currently) involves a URL redirect.

**Comment by HenrikBengtsson** [comment]
Forgot to say, I finally got the build environment up so I could run:

```
./configure && make 
make check
```

which passes with all OK.

PR #593, opened 2026-04-09
Add libpoppler recipe for pdftools R package support

### LINKED ISSUE #575: add libpoppler for pdftools
I would love to have `pdftools` available on `webr` if at all possible (ref https://github.com/ropensci/pdftools/issues/139).  Thanks so much for your consideration and the absolutely brilliant `webr`.  If I had any clue what I was doing, I would happily volunteer, but unfortunately this is currenlty outside of my abilities.

lpmi-13:
Add a WebAssembly build recipe for libpoppler (v24.12.0) with the C++ wrapper (poppler-cpp) enabled, which is required by the pdftools R package.

Closes r-wasm/webr#575

NB: I'm primarily raising this PR to see if it builds and passes CI okay, since I'm on holiday and only have access to a mobile device, but just found out about this project and excited to contribute.

I also didn't notice any language around LLM contributions in the README or Code of Conduct, but since this code was generated via Opus 4.6, I wanted to call that out just in case the maintainers would prefer no AI contributions. I will happily close this PR if that's the case.


lpmi-13:
realised that since I had committed from my mobile device, I wasn't able to sign the commit with my ssh key, so I've done that and force pushed.

lpmi-13:
I also attempted to run through `./configure && make`, but my laptop crashed about halfway through, so I didn't get to the end of compilation. I'm happy to do that if it would be helpful (I don't want to give the maintainers more work to do :sweat_smile: )

georgestagg:
No worries, building the WebAssembly libraries from source is a messy endeavour, as it requires building a bunch of heavy compiler dependencies that have been customised for our use.

I've pushed a commit to build all libraries in CI as part of the PR testing, it just takes a very long time. If the build-libs workflow returns green in an hour or so, we can merge.

PR #595, opened 2026-04-15
Update libsodium to v1.0.22

georgestagg:


jeroen:
I think the CI for pull requests does not build recipes?

georgestagg:
`make all` for the recipes will be run once merged to `main`.

We should add a rule in the PR workflow to do a full `make all` if any files under `libs` are touched.

PR #596, opened 2026-04-15
Test full curl request

jeroen:


jeroen:
OK seems like this works now, great!

georgestagg:
Awesome, I'll merge in this test now then!

PR #598, opened 2026-04-16
Port libuv from fs package

jeroen:
This literally copies the custom libuv from https://github.com/r-wasm/fs/tree/webr


georgestagg (approved):
Great, thanks!

PR #609, opened 2026-06-08
Add recipe for protobuf

jeroen:
This will unblock a heap of packages, such as protolite, Rprotobuf, tfevents, otelsdk, cld3


georgestagg (approved):
LGTM, thanks!

